PR #15, opened 2025-03-03
CoSMIC docker image added

Adnan525:
Changes -

- Added Dockerfile for CoSMIC.
- run_chatbot.sh has a dynamic host IP and a shared volume to share data which were previously mounted directly in the container
- pipelines/start_docker.sh added which starts the pipeline server. Host IP is assigned dynamically.
- Updated Readme.


README.md
```diff
@@ -64,39 +104,48 @@ jwt==1.3.1
...
+To use ["code generation and evaluation"](modules/code_generation/code_generation.py) service, users need to install [docker](https://docs.docker.com/engine/install/):
+
 ```bash
 apt install docker.io
 ```
 
+### For Docker Installation
+
+- Docker must be installed on your local machine
+- An OpenAI API key is required for full functionality
```
carlosclaiton: don't you consider a user using ollama that the system will be full functional?

README.md
```diff
@@ -1,34 +1,74 @@
...
+  --network=host \
+  -v /absolute/path/to/your/.env:/app/.env \
+  -v /var/run/docker.sock:/var/run/docker.sock \
+  --privileged \
+  opensicbr/cosmic:demo
+```
+
+**Note**: While the Docker setup offers a streamlined environment for testing, it grants direct access to the host network and is **not recommended for production environments**.
+
 ## Requirements
```
carlosclaiton: don't we need nay more requirements

README.md
```diff
@@ -64,39 +104,48 @@ jwt==1.3.1
...
 - [FEN generation given a sequence of moves](src/modules/chess_genfen.py)
 - [Chain-of-Thought generation for next move prediction](src/modules/chess_gencot.py)
 
+## Getting Started
 
-## Get Started
+### Web-based Chatbot
 
-### [General User] Chatbot
+We provide a website-based chatbot for interaction between users and OpenSI-CoSMIC.
```
carlosclaiton: the docker version we are leaving as option 2, already has the front end to it, should this section be named as getting started?

carlosclaiton (changes requested):
I know that is more semantics, but it is better to get this write

README.md
Adnan525: we don't need any more requirements, the docker image has everything installed in it. Looking into Ollama now.

README.md
```diff
@@ -1,34 +1,74 @@
...
+Pull the latest CoSMIC image:
+
+```bash
+docker pull opensicbr/cosmic:demo
+```
+
+Create an `.env` file to store your OpenAI API key:
+
+```bash
+echo "OPENAI_API_KEY='your_api_key'" > /path/to/save/.env
```
zwxu064: set an option here as openai_api_key is not required for ollama users.

README.md
zwxu064: I agree with Carlos for users has no intension to use openai api but only local models from Ollama.

Dockerfile
zwxu064, line 1: If Line 30 is not used, delete it.
Do not leave commented commands.

scripts/chatbot/cosmic.sh
```diff
@@ -0,0 +1,2 @@
+ollama serve > ollama.log 2>&1 &
+bash /app/run_chatbot.sh &
```
carlosclaiton: That means you have ollama native installed inside of the docker container, it is not using Ollama docker, correct.

run_chatbot.sh
```diff
@@ -1,32 +1,51 @@
...
-    --mount type=bind,source="${SCRIPT_DIR}/.env",target="/app/backend/.env" \
+    -e ENV_CONTENT="$ENV_CONTENT" \
     "$image_name"
 
 docker image prune -f
 
+# run the pipeline script
 cd ${SCRIPT_DIR}/modules/chatbot/pipelines
-bash start.sh
+bash start_docker.sh
```
carlosclaiton: Have you modified this .sh file, it is not showing on the modifications. Maybe you just is starting a file already there?

run_chatbot.sh
```diff
@@ -1,32 +1,51 @@
...
 fi
 
+# build the image
 docker build -t "$image_name" ${SCRIPT_DIR}/modules/chatbot
-docker stop "$container_name" &>/dev/null || true
 docker rm "$container_name" &>/dev/null || true
 
-docker run -d -p "$host_port":"$container_port" \
-    --add-host=host.docker.internal:host-gateway \
+# sharing the host network so port forwarding is not needed
```
carlosclaiton: can we add an env variable to differentiate when running it as develpment mode or production? that will avoid commenting out part of the code and having to modify it later.

scripts/chatbot/cosmic.sh
carlosclaiton: why did not modified the run_chatbot.sh to add line 1 into it directly?

README.md
carlosclaiton: we do if not using docker, that is why you need to let people to know the requirements, and if people want to build the docker it will need the requirement files, am i right?

carlosclaiton (changes requested):
it is getting there, nice work

scripts/chatbot/cosmic.sh
Adnan525: yes, not using ollama docker container, it will take a bit of time and I will try this first with pycapsule.
did you mean to add the ollama installation to run_chatbot.sh? if yes, then that will work but I just prefer to do this via Dockerfile since it's a program installation along with docker, curl etc.

run_chatbot.sh
Adnan525: new file

carlosclaiton: I don't think you push this file then.

run_chatbot.sh
Adnan525: probably but for the time being, I don't know how to have this working without sharing the whole network. I suppose for a demo, that's the best we can do.

carlosclaiton (approved):
I am happy to merge

PR #24, opened 2025-05-13
Chat history production

Adnan525:
Only adds "user" and "assistant" content as chat history.
Check utils/chat_history.py for functionality


Adnan525:
the solution may not work with rag questions, I have converted it to a draft, will have a look soon

Adnan525:
Tested with conversation history pairs and RAG.

PR #26, opened 2025-05-14
Readme and other minor modifications.

zohaibhammad:


README.md
```diff
@@ -10,11 +10,33 @@ This is the official implementation of the Open Source Institute-Cognitive Syste
 
 ## Installation
 
+### Pre-requirements
+
+Before proceeding with the installation, ensure that the following tools are installed on your local machine:
+
+1. **Git**: Used for cloning repositories. You can install it by following the [official Git installation guide](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git).
```
carlosclaiton, line 17: Git is required only if not the quick install, maybe we need to change the order.

README.md
```diff
@@ -44,31 +66,47 @@ bash docker_compile.sh
...
 # For users using HTTPS
-git clone --recursive https://github.com/TheOpenSI/CoSMIC.git
+git clone https://github.com/TheOpenSI/OpenWebUI-CoSMIC.git
 ```
 
-2. For chess-game queries, users need to [download](https://stockfishchess.org/download/linux/) the Stockfish binary file (stockfish-ubuntu-x86-64-avx2 for Linux) and store it at the default location: `third_party/stockfish/stockfish-ubuntu-x86-64-avx2`.
+**Note**: Ensure that both repositories are cloned into the same directory to maintain compatibility.
+
+
+3. (Optional) For chess-game queries, users need to [download](https://stockfishchess.org/download/linux/) the Stockfish binary file (stockfish-ubuntu-x86-64-avx2 for Linux) and store it at the default location: `third_party/stockfish/stockfish-ubuntu-x86-64-avx2`.
```
carlosclaiton, line 93: Is not it already part of the repo? why do we need that information here

zohaibhammad: It is. As it was already there in the readme, I kept it. But I'll remove this part.

PR #30, opened 2025-05-19
Updated Readme, docker-compose and api.py file

zohaibhammad:


api.py
carlosclaiton, line 1: have you checked if this will not break the code, as it was working before

docker-compose.yaml
```diff
@@ -26,13 +26,12 @@ services:
...
-    # build:
-    #   context: ./../OpenWebUI-CoSMIC
-    #   dockerfile: Dockerfile
-      # args:
-      #   OLLAMA_BASE_URL: '/ollama'
+    build:
+      context: ./../OpenWebUI-CoSMIC
+      dockerfile: Dockerfile
+      args:
+        OLLAMA_BASE_URL: '/ollama'
```
carlosclaiton, line 33: I think this is required only when we are using ollama container, which is commented by now. Adnan is working on the version that uses Ollama as container.

carlosclaiton (changes requested):
Can you please redo the PR to https://github.com/TheOpenSI/CoSMIC/tree/production_pre-release branch instead, we should from now on to do the modifications in this branch and when we finish the whole sprint we then merge into production changing the version. I forget to mention this new routine.

api.py
zohaibhammad: Yes, this is how our project ran all the time before it was changed.

docker-compose.yaml
zohaibhammad: Yes, but it can remain as it is because it doesn't affect the functionality.

PR #32, opened 2025-05-20
Ollama integration

Adnan525:
Added new ollama class and separated the LLM  base class.
Implemented with an ollama client instance

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -143,7 +143,7 @@ services:
 
 
 volumes:
-  # ollama: {}
+  ollama: {}
   open-webui: {}
   pipelines: {}
   volume_uploads: {}
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,14 +1,18 @@
 services:
-  # ollama:
-  #   image: ollama/ollama:${OLLAMA_DOCKER_TAG:-latest}
-  #   container_name: ollama
-  #   volumes:
-  #     - ollama:/root/.ollama
-  #   pull_policy: always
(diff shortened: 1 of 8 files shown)


docker-compose.yaml
```diff
@@ -1,14 +1,18 @@
...
+  ollama:
+    image: ollama/ollama:latest
+    container_name: ollama
+    deploy:
+      resources:
+        reservations:
+          devices:
+            - driver: nvidia
+              count: all
+              capabilities: [gpu]
```
carlosclaiton, line 11: doo you need that, will it still work at a cpu?

src/services/llms/llm.py
```diff
@@ -563,36 +321,36 @@ def get_openai_key(self):
 
 # =============================================================================================================
```
carlosclaiton, line 323: this class was already there before? it is confusing to know what is new

carlosclaiton (changes requested):
we need to make sure it will run in a normal laptop with this implementation as well.

docker-compose.yaml
Adnan525: we will have to test this and if it does not work, we will put this information in readme

src/services/llms/llm.py
Adnan525: the lllm.py file had LLMBase and Ollama class before.
LLMBase class had been moved to a new file, commented out Ollama class and added new implementation.
Only addition to that file would be 2 import statements for LLMBase and Ollama

PR #34, opened 2025-05-26
Zohaib dev

zohaibhammad:
- API clean-up
- OpenAI API key handling in .env
- Config update issue

### DIFF
--- a/api.py
+++ b/api.py
@@ -8,6 +8,7 @@
 # =======================================================
 
 from datetime import datetime
+import dotenv
 from fastapi import FastAPI, Request, File, UploadFile, HTTPException
 from fastapi.middleware.cors import CORSMiddleware
 from src.opensi_cosmic import OpenSICoSMIC
@@ -18,6 +19,7 @@ from zoneinfo import ZoneInfo
 from typing import Optional
 
 from utils.chat_history import build_context_from_messages
+from utils.log_tool import set_color
 from utils.statistics import update_statistic_per_query
 
 app = FastAPI()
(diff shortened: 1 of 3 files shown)


api.py
```diff
@@ -38,117 +40,29 @@ class CosmicAPI(BaseModel):
...
-            statistic_dict["query_count"] = 1
-            update_statistic_table(statistic_dict)
+openai_api_key = os.environ.get("OPENAI_API_KEY", "")
+
 
 def update_openai_key():
     global openai_api_key
     # Set up OPENAI_API_KEY globally through root's .env.
-    openai_api_key = os.environ.get("OPENAI_API_KEY", "0p3n-w3bu!")
+    openai_api_key = os.environ.get("OPENAI_API_KEY", "")
```
carlosclaiton: You are getting this insed of the function and also outside of the function.
Also you just moved function that are not api calls from the api, this should be implemented in another file.

api.py
```diff
@@ -205,6 +119,24 @@ class ConfigUpdateForm(BaseModel):
 
 openai_api_status = opensi_cosmic.check_openai_key()
 
+def rebuild_cosmic():
```
carlosclaiton, line 110: This also should be outside of the API.
And I also would appreciate you commenting the function. For example:

def rebuild_cosmic():
     """  This funciton will ... due to ... """

carlosclaiton (changes requested):
In general comment and questions, all those classes you are defining in the api.py should it be define somewhere else and call here? What is the best practice? I would put all the schemas (that uses baseModel) in a different file to keep the api.py as clean as possible.

https://github.com/zhanymkanov/fastapi-best-practices

carlosclaiton:
I also find the bug when trying to save the config update:

Screenshot 2025-05-27 at 8 20 21 am

carlosclaiton:
In my local test it did not update the OPENAPI_KEY on the .env when I update this information on the frontend config page.
@zohaibhammad

PR #36, opened 2025-05-28
Openai key validation

zohaibhammad:
- Implemented OpenAI API key validation.
- Reverted the multi-select code.

### DIFF
--- a/api.py
+++ b/api.py
@@ -8,6 +8,7 @@
 # =======================================================
 
 from datetime import datetime
+import dotenv
 from fastapi import FastAPI, Request, File, UploadFile, HTTPException
 from fastapi.middleware.cors import CORSMiddleware
 from src.opensi_cosmic import OpenSICoSMIC
@@ -18,6 +19,7 @@ from zoneinfo import ZoneInfo
 from typing import Optional
 
 from utils.chat_history import build_context_from_messages
+from utils.log_tool import set_color
 from utils.statistics import update_statistic_per_query
 
 app = FastAPI()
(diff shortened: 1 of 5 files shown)


api.py
```diff
@@ -38,117 +42,16 @@ class CosmicAPI(BaseModel):
...
     global openai_api_key
-    # Set up OPENAI_API_KEY globally through root's .env.
-    openai_api_key = os.environ.get("OPENAI_API_KEY", "0p3n-w3bu!")
+    openai_api_key = os.environ.get("OPENAI_API_KEY", dotenv.dotenv_values(".env").get("OPENAI_API_KEY", ""))
+    if not openai_api_key:
+        print(set_color("warning", "OPENAI_API_KEY is required in .env or environment variables."))
+
+
+# Initialize the OPENAI_API_KEY on startup.
+update_openai_key()
```
carlosclaiton, line 55: this will raise the following warning:

[Warning] OPENAI_API_KEY is required in .env or environment variables.

if we are ysing the system without OpenAI

carlosclaiton (changes requested):
and config page is not laoding, but I tested the api and it is working, problem probably is on the OpenWebUI-CoSMIC.

Can you pleaseMove fucntion: update_openai_key() and rebuild_cosmic() to become part of OpenSICoSMIC class.

That way it can initiate cosmic at the first time we initiate the class and call this funcion everytime we need to update the config file.

zohaibhammad:
@carlosclaiton This PR is updated. You can merge it now.

PR #37, opened 2025-06-03
CORS

zohaibhammad:
feat: Update CORS configuration to allow specified origins from environment variables in CoSMIC and Open-WebUI

### DIFF
--- a/api.py
+++ b/api.py
@@ -26,9 +26,22 @@ from utils.statistics import update_statistic_per_query
 
 app = FastAPI()
 
+# To test CORS_ALLOW_ORIGIN locally, you can set something like
+# CORS_ALLOW_ORIGIN=http://localhost:5173;http://localhost:8080
+# in your .env file depending on your frontend port, 8080 or 5173 in this case.
+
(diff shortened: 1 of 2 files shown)


api.py
```diff
@@ -26,9 +26,22 @@
 
 app = FastAPI()
 
+# To test CORS_ALLOW_ORIGIN locally, you can set something like
+# CORS_ALLOW_ORIGIN=http://localhost:5173;http://localhost:8080
+# in your .env file depending on your frontend port, 8080 or 5173 in this case.
+
+CORS_ALLOW_ORIGIN = [
+    "http://localhost:8080",  # Open-WebUI production server
+    "http://localhost:5173",  # Open-WebUI development server
```
carlosclaiton, line 35: don't we need to add here the ollama, postgress port as well?

carlosclaiton (approved):
you can merge this, but please answer my question, just for curiosity.

api.py
zohaibhammad: Nope. There should only be addresses that will access our APIs. None other than that.

PR #44, opened 2025-08-04
Ollama model pulling issue

Adnan525:
Model pulling issue fixed, will restart the model pulling at set thresholds.
Pulls model when the LLM object is instantiated, not when CoSMIC receives a query.


marslanshaukat (approved):
Tested on cleaned environment and found no issue. 

Thanks.

PR #46, opened 2025-08-08
Cosmic 81

Adnan525:
- PyCapsule integration
- Updated readme
- Removed contents of .env, works fine

image


README.md
```diff
@@ -89,8 +89,9 @@ cd CoSMIC
 
 6. Now you can build from your local clone using the command below:
 ```bash
-docker compose up -d --build
```
carlosclaiton, line 92: have to update the first method of installation as well. 
But modify the script to include the wget
Can you also please check the default .env

carlosclaiton (changes requested):
also fix the .env issue

carlosclaiton (approved):
Well done, it is working as expected.

PR #53, opened 2025-10-13
Multi select backend

josephalanbenny:
Updated service detection logic and configuration handling.


carlosclaiton:
@josephalanbenny, it should be a merge to dev not to production. As production is only when everything has pass all the tests

PR #69, opened 2026-03-18
Models management

manilesrun:
Models management apis - view, pull, and delete

### DIFF
--- a/scripts/chatbot/cosmic.sh
+++ b/scripts/chatbot/cosmic.sh
@@ -6,4 +6,4 @@ ollama serve > ollama.log 2>&1 &
 # bash /app/scripts/chatbot/run_chatbot_docker.sh
 
 # https://www.uvicorn.org/settings/#configuration-methods
-uvicorn api:app --host 0.0.0.0 --port 3000  --reload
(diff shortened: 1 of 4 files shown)


scripts/chatbot/cosmic.sh
```diff
@@ -6,4 +6,4 @@ ollama serve > ollama.log 2>&1 &
 # bash /app/scripts/chatbot/run_chatbot_docker.sh
 
 # https://www.uvicorn.org/settings/#configuration-methods
-uvicorn api:app --host 0.0.0.0 --port 3000
\ No newline at end of file
+uvicorn api:app --host 0.0.0.0 --port 3000  --reload
```
carlosclaiton: We need to make sure we change this when we create the production

carlosclaiton: Write a #todo here to remember that we should not use --reload

carlosclaiton (approved):
Just make sure we leave something on that code that indicates the changes we need to do on the future. 

You can merge into development branch, it is downloading, deleting and updating the models as expected.

PR #71, opened 2026-03-18
Temporary uploading `data` folder and `config_updated.yaml` file for CoSMIC to run [COSMIC-113]

FaceWithDark:
The way CoSMIC code finding the folders and files it needs to run is in a very questionable style. Therefore, we’ll public these data for now (nothing too sensitive in these actually, but it’s polluting the repo size).


FaceWithDark:
LGTM. I'll merge it now.

PR #72, opened 2026-03-18
Update backend docs addressing the temporary `todo.env` file for OpenAI API key [COSMIC-114]

FaceWithDark:
CoSMIC check for OpenAI API key for using models from OpenAI. Only one file successfully found this while the rest is not, but if the file not there then an FastAPI build error would just appear.


FaceWithDark:
LGTM. I'll merge it now.

PR #76, opened 2026-03-19
Academic governance

carlosclaiton:
This allows for automatic detection of system prompts per service. It forms the base for the capability of adding new services or agents in cosmic.


Adnan525:
See pull request #79

PR #78, opened 2026-03-19
Merge working solution `lifespan()` async func over to **Development** branch [COSMIC-127]

FaceWithDark:
Update `requirements.txt` due to the new `psycopg` library added so that PostgreSQL understand the connection request from FastAPI (or Python) is done through an adapter that "talk" the same db language.


FaceWithDark:
LGTM. I'll merge this PR.

PR #86, opened 2026-03-21
Remove `scripts` directory (only needed on the big "tree") [COSMIC-134]

FaceWithDark:
These scripts are not needed in the new repo as all util scripts & files will go into `utils` directory instead.


FaceWithDark:
LGTM. I'll merge this PR.

PR #87, opened 2026-03-21
Remove and adjust `.git*` suffix files [COSMIC-135]

FaceWithDark:
We only need to keep those configurations that are relevant to BE development.


FaceWithDark:
LGTM. I'll merge this PR.

PR #90, opened 2026-03-21
Sub-commit for [COSMIC-137] (Part 2)

FaceWithDark:
- Update naming for each services to prevent using images and volumes from the "tree" when debugging on multiple repos at once.
- Remove the need for OpenAI API key as that is only needed from the CoSMIC repo, not the new one.
- Remove logic for `/cosmic` POST request call as that is handle from the CoSMIC repo instead. We only try to call from that repo through an agreed open port/domain/proxy.
- Remove AI/ML packages as that is only needed from the CoSMIC repo, not the new one.


FaceWithDark:
LGTM. I'll merge this PR.

PR #91, opened 2026-03-21
Simplify Docker Compose configurations as this's now a mono repo [COSMIC-137]

FaceWithDark:
As title said.


FaceWithDark:
LGTM. Need to do minor git conflict from terminal but resolved.

PR #94, opened 2026-03-22
Remove other normal endpoints except `/cosmic` [COSMIC-141]

FaceWithDark:
This is pretty clear as those old endpoints should be routed from FE instead. Technically those endpoints are only needed to render diferent pages, no need any data from our BE.


FaceWithDark:
LGTM. I'll merge this PR. From this point, no more tasks are needed for migration from other branches anymore.

PR #97, opened 2026-03-23
migrate Cosmic to use UV [COSMIC-129]

FaceWithDark:
I have tested on my end. Migrate to `uv` have no issue, got the models installed, test call to `/cosmic` endpoint and got response back from installed model.


FaceWithDark:
For the changes on `langchain_text_splitters` package. Previously, they did a typo on the package name so on newer version of that package, they updated to fix the naming. Functional-wise are the same, just the naming is different.

FaceWithDark:
I've also add the explicit `None` type-hint so that the linter not screaming at me, and it's more explicitly correct as well.

api.py
```diff
@@ -378,4 +389,98 @@ async def process_cosmic(data: CosmicAPI):
 
 
 # Models APIs - smanile
+@app.get("/models")
```
carlosclaiton: This is no longer required. Can you please do a git pull to ensure you are using the most up-to-date code?

carlosclaiton (changes requested):
Manile has moved part of the api.py to a folder backend/routers, but you did not pull to properly merge the most recent code. 

Can you please ensure the code is up-to-date?

api.py
FaceWithDark: No worries, I just pulled it now (commit 17c820a).

carlosclaiton (approved):
You can merge now. I is running normally on my machine, we do have a  new warnin

BertModel LOAD REPORT from: thenlper/gte-small
Key                     | Status     |  | 
------------------------+------------+--+-
embeddings.position_ids | UNEXPECTED |  | 

Notes:
- UNEXPECTED    :can be ignored when loading from different task/architecture; not ok if you expect identical arch.
Adding documents to vector database...
Reconstruct OpenSICoSMIC due to changed configs.
Loading weights: 100% 199/199 [00:00<00:00, 5209.27it/s]
BertModel LOAD REPORT from: thenlper/gte-small
Key                     | Status     |  | 
------------------------+------------+--+-
embeddings.position_ids | UNEXPECTED |  | 

Notes:
- UNEXPECTED    :can be ignored when loading from different task/architecture; not ok if you expect identical arch.

I am not much concerned as we are about to change the RAG embedding model soon anyway.

FaceWithDark:
No worries, I'll get this PR merged now.

PR #99, opened 2026-03-27
chore: move old endpoints to a new file

manilesrun:
Cosmic-148:  cosmic api to routers folder

Hi Prof, 

I have moved the old endpoints to default_apis.py. Everything is the same no change just move file. You can test the apis using local:3000. They will still work fine. If not, please let me know.

Later on, if we want to improve cosmic, we can just upgrade the endpoint to V1 to save chat history. So, we can put the new cosmic_V1 file in the path: /routers/cosmic_v1 

this way we will not change any of the old code, we just change the endpoint and the frontend just use the new endpoint. If the new endpoint not working, we can always roll back to the previous version.


carlosclaiton (approved):
you can merge this PR. Thanks

PR #102, opened 2026-03-30
Cosmic 173

Adnan525:
document reference fixed, meta-data only contains document title

### DIFF
--- a/scripts/configs/config.yaml
+++ b/scripts/configs/config.yaml
@@ -9,7 +9,8 @@ is_quantized: False
 seed: 0  # seed of response generation by base LLM.
 doc_directory: "/app/data/docs"  # directory containing .pdf documents to build the system's vector database.
 document_path: []  # a string or a list of .pdf to add context to the vector database.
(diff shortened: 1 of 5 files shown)


scripts/configs/config.yaml
```diff
@@ -9,7 +9,8 @@ is_quantized: False
 seed: 0  # seed of response generation by base LLM.
 doc_directory: "/app/data/docs"  # directory containing .pdf documents to build the system's vector database.
 document_path: []  # a string or a list of .pdf to add context to the vector database.
-service: 3  # service selection: -1 [auto], 0 [chess], 1 [update vector database], 2 [code generation], 3 [general QA]
+service: 4  # service selection: -1 [auto], 0 [chess], 1 [update vector database],
```
carlosclaiton: on development it should be -1

carlosclaiton (approved):
You can merge, but please set the config file to -1 for the query analyser, as we need always to check its capability.

Issue #103, opened 2026-03-31
OpenAI library


Adnan525:
Newer OpenAI models will require updating the OpenAI python library, a few old parameters are not supported in the new response/completion endpoints. Test with new openai models and update the library if required.

Issue #105, opened 2026-04-01
LLM Service Name


Adnan525:
At LLM base __call__ the parameters are - 
def __call__(
        self,
        question: str,
        context: dict = {},
        service: str = ""
    ):
and at Ollama - 
    def __call__(self,
                 question: str,
                 context: Union[str, Dict[str, Any]] = {},
                 serivce_name: str = "",
                 ) -> tuple[str, str]:
fix the typo and make the calls consistent

PR #106, opened 2026-04-01
service name fixed, openai library updated and api call fixed

Adnan525:
Closes issue #103 and #105


carlosclaiton (changes requested):
It is not correct as the Development branch does not use requirements anymore; it has been migrated to uv instead of pip install. 

Can you ensure you use the most up-to-date development branch to apply the fix?

Adnan525:
tested with UV, works fine now. no changes required, removed requirements.txt

src/services/llms/Ollama.py
```diff
@@ -134,9 +133,3 @@ def quit(self) -> None:
         Clean up any resources. No specific cleanup needed for API-based implementation.
         """
         pass  # No resources to clean up for REST API implementation
-    
-# if __name__ == "__main__":
```
carlosclaiton, line 138: this part of the code is supose to prevent this c ode from running when the script is imported and provide a clean entry point for the program, are you sure you want to remove it.

carlosclaiton (changes requested):
It does not work with the frontend; the response is not going to the frontend. I could not find the reason.

src/services/llms/Ollama.py
Adnan525: yes that's fine

Adnan525:
Typo "serivce_name". Please check now

carlosclaiton (approved):
That seems to be working now, thanks.

Issue #109, opened 2026-04-09
Document Embedding and Vector DB loading


Adnan525:
Document embedding and loading into the vector database is exceptionally slow when done from the container. To recreate, start up the application from the Docker Compose and run the demo.py from the RAG_AG_BM for local run.

PR #110, opened 2026-04-09
Added gpu support to cosmic container

Adnan525:
Fixes issue #109 
Please test with all 8 documents, did not add those since the Development branch does not have those docs.


carlosclaiton (changes requested):
I tried but I failed, for some reason it does not work when we ask questions about academic governance

Adnan525:
Carlos will send a new pull request with stats fix.

PR #112, opened 2026-04-10
bug:fix statistic file that was causing api to crash and issue 109

carlosclaiton:
Fixed the issue with the statistics
using uvicorn instead of fastapi CLI, to avoid the bad printing
Fixed issue #109


Adnan525 (approved):
Tested with Carlos, stats data type changed, tested with 8 pdfs and added uvicorn.

PR #117, opened 2026-04-17
CoSMIC read services from DB [COSMIC-120] (Old)

FaceWithDark:
**Re: [COSMIC-120 Ticket](https://opensi.atlassian.net/browse/COSMIC-120?atlOrigin=eyJpIjoiN2IxMTRiNTI5ZTA4NGI4ZWIwYzAxYjg1MzFjMWZjYTgiLCJwIjoiaiJ9)**

Extra update along ticket description information:
- Add new `types` directory for consistent directory structure with [`COSMIC-DB` repo](https://github.com/TheOpenSI/COSMIC-DB/tree/dev/types). Within this directory, there's a new `query_analyser.py` file that mainly for resolving faisely error-level diagnostic trigger from `pyright`.
- Modify regex string provided in each `.search()` method to use **r-string** format as explained [here](https://docs.python.org/3/library/re.html#raw-string-notation).
- Added documentation for each new function/class following [Sphinx format](https://www.sphinx-doc.org/en/master/index.html) - Python's most used documentation generator tool and format. It's even used by [Python documentation page itself](https://devguide.python.org/documentation/help-documenting/).
- For verbose option enabled, I used a mix of built-in `print()` method and [`pprint` library](https://docs.python.org/3/library/pprint.html) for pretty format rather than just `print()` method itself, which can be hard to read for non-simple type data (e.g., dict, nested data type, etc).


src/services/qa.py
```diff
@@ -269,3 +284,92 @@ def __call__(
...
+            Example: {0: "memory", 1: "code_generation"}
+
+        Raises:
+            HTTPException: With status code 500 if any connection error occurs
+                (ConnectError, ConnectTimeout) or other unexpected exceptions.
+
+        Example:
+            >>> services = obj._get_services_name(verbose=True)
+            >>> services[0]  # First service name
+            'academic_governance'
```
carlosclaiton: This code will not function as it should; the file name at /src/services/llms/prompst needs to match the service name to retrieve the correct system prompt.

src/query_analyser/query_analyser.py
```diff
@@ -39,25 +54,29 @@ def __init__(
...
-            "2": "generate or improve a code or answer a question in order to generate or improve a code",
-            "3": "answer a question or provide a reasoning, which cannot be achieved by the other services",
-            "4": "Answer question about Academic Governance"
+
+        # NOTE:
+        # Due to the way current logic checking for selected service after
+        # receiving the response from LLM based on Query Analyser's refined
+        # question from user query, we've to do this kind of workaround until
+        # it get updates to handle 'int' type correctly (which it should be).
+        services_desc:              dict[int, str] = self._get_service_desc()
```
carlosclaiton: In this function, it is not clear to me if it is returning only the services that have a status of true. 
It is important that the query analyser is fed only with available services set by the administrator.

carlosclaiton (changes requested):
@FaceWithDark, I don't see how you are handling:
- only services with true status
- what happens if the list is empty?
- name of the file for the prompt should match with the name of the service.

carlosclaiton:
I have tested, first ask cosmic how can you help me (it is taking much longer to generate the anser)

then go to the database and 

UPDATE services
SET status = FALSE
WHERE status = TRUE
  AND id IN (1, 2);
  
  It should disable the services, ask the same question in a new chat, and clear the chat history; it gives the same answer as before. It should read from the database and know which services are available.

FaceWithDark:
> @FaceWithDark, I don't see how you are handling:
> - only services with true status
> - what happens if the list is empty?
> - name of the file for the prompt should match with the name of the service.
> 

I'll take a look into these validation. The way I wrote current code will run normally under perfect condition (there is/are data in the db, and all services is active by default).

I don't really get what you meant for the last point. Can I get some more info on it? (an example would do).

carlosclaiton:
> > @FaceWithDark, I don't see how you are handling:
> > 
> > * only services with true status
> > * what happens if the list is empty?
> > * name of the file for the prompt should match with the name of the service.
> 
> I'll take a look into these validation. The way I wrote current code will run normally under perfect condition (there is/are data in the db, and all services is active by default).
> 
> I don't really get what you meant for the last point. Can I get some more info on it? (an example would do).

In the qa.py it gets the service name to allow the system_prompt.py to check if there is a txt file with the same name of the service name in the folder /src/services/llms/prompts. At moment the file is called AcademicGovernance.txt and the service name is called academic_governance. The system will use general question system prompt if a user ask question about academic governance if the name does not match. 


I also think the way to solvet the status you should have a dedicated endpoint that returns only services with status true.

FaceWithDark:
Based on the reviews, I think we have to froze this PR as some extra implementations are needed in `COSMIC-DB` repo in order to be able to resolve these reviews.

src/query_analyser/query_analyser.py
FaceWithDark: I've opened a PR in `COSMIC-DB` repo that would resolve this review. I'll update the code in this PR when that one is merged to `dev` branch (Ref: [PR #26](https://github.com/TheOpenSI/COSMIC-DB/pull/26)).

src/services/qa.py
FaceWithDark: I've tested this with your instruction in comment below with the new commit 3c4ed51. It's perfectly capture my question and response with content it found from the `academic_governance.txt` file:
```txt
what is academic governance? --> <response>
```

FaceWithDark:
**UPDATE**:

> [!IMPORTANT]
> Commit 2ab50f1 attempt to match compose file naming convention like other modular repos (`compose.yaml`). This will help (visually) finding the right compose file across repos as it's in a consistent naming, and also simply the unified compose file in [`COSMIC_Docker` repo](https://github.com/TheOpenSI/CoSMIC_Docker).

src/query_analyser/query_analyser.py
FaceWithDark: Code is updated and tested (commit [bf828f4](https://github.com/TheOpenSI/CoSMIC/pull/117/commits/bf828f4039f4835745259c7fdbc52b5d48fac115))

FaceWithDark:
**UPDATE:**
With commit ea4a26d pushed (addressing the case when the list of active services is empty), I think is PR is good to for another review before merging now. I've also request review from @Adnan525 to boost up this process a bit, as per our conversation before. However, I did put a `TODO` note there to see whether both of you wanted me to handle this case that way or the other way around

scripts/configs/config.yaml
```diff
@@ -5,21 +5,28 @@
...
+# d) 2  [code generation]
+# e) 3  [general QA]
+# f) 4  [Academic QA]
+service: -1
 
 query_analyser:
-  llm_name: "ollama:qwen2.5:7b"
-  is_quantized: False
+  llm_name: "ollama:llama3.3:70b"
+  is_quantized: true
```
carlosclaiton: why using a very large model here? are you trying to test for quantiszed models?

src/query_analyser/query_analyser.py
```diff
@@ -39,76 +54,98 @@ def __init__(
...
+                for (str_id, desc) in services_desc.items()
+            }
 
-        # Get the number of services.
-        self.num_services = len(self.services)
+            # Set chess subservices.
+            self.chess_subservices_desc: dict[str, str] = {
+                "0.0": "predict next move given a chess FEN",
+                "0.1": "predict next move given a sequence of moves"
+            }
```
carlosclaiton: why is this subservices required now?

scripts/configs/config.yaml
FaceWithDark: I was just finding a newer model than the one we currently used in `production` branch (within the same model family and capability as well). I did tested asking some questions on this model with quantise enabled and didn't find any issues so far except it runs a bit slow on my desktop using Windows (which means I could only use CPU).

src/query_analyser/query_analyser.py
FaceWithDark: It's part of the old codebase so I just keep it to make sure the old logic still running

Adnan525 (changes requested):
Service selection is faulty. Whenever I asked something along the line of "who are you" or "how can you help me" it either forwards the query to pycapsule or fails with 500 internal server error. It worked a few times but mostly routed to pycapusle. Additionallty something is wrong with -1 service. see attached screenshots 

Image
Image

carlosclaiton:
I also have found several issues, it seems the query analyser not take into consideration the list of available services, as you can see on the images bellow, it chose a service that is not available and also replay as if all services ara available.
Screenshot 2026-05-06 at 12 26 18 pm

FaceWithDark:
> [Adnan](https://github.com/TheOpenSI/CoSMIC/pull/117#pullrequestreview-4230437594):
> Service selection is faulty. Whenever I asked something along the line of "who are you" or "how can you help me" it either forwards the query to pycapsule or fails with 500 internal server error. It worked a few times but mostly routed to pycapusle. Additionallty something is wrong with -1 service. see attached screenshots
> 
> Image Image


> [Carlos](https://github.com/TheOpenSI/CoSMIC/pull/117#issuecomment-4384652086):
> I also have found several issues, it seems the query analyser not take into consideration the list of available services, as you can see on the images bellow, it chose a service that is not available and also replay as if all services ara available. Screenshot 2026-05-06 at 12 26 18 pm


Hi, sorry for being inactive for the last 2 days. I was focusing on finish the reports so I wasn't aware of the raised issues.

It seemed to me that those problem are within the core code than my changes to dynamic call active services from our db.

I got some freetime today so I'll take a deeper look in the core code to see if it's the case.
> [!NOTE]
> We might have to upgrade this PR into a bigger Jira ticket scope (with subtasks) if the issue landed on core code

carlosclaiton:
This only happens with this branch; it has to be something related to these modifications.

FaceWithDark:
> This only happens with this branch; it has to be something related to these modifications.

I'm just wondering if it's the issue with the new `/cosmic` endpoint @manilesrun created since I'm asking questions that would trigger the capable service in our old `/cosmic` endpoint. It does use the right service (I didn't deactive any services so all of them are availalbe for use, similar to the pre-populated script).

image

image

image

image

FaceWithDark:
> > This only happens with this branch; it has to be something related to these modifications.
> 
> I'm just wondering if it's the issue with the new `/cosmic` endpoint @manilesrun created since I'm asking questions that would trigger the capable service in our old `/cosmic` endpoint. It does use the right service (I didn't deactive any services so all of them are availalbe for use, similar to the pre-populated script).
> 
> image image image image

When I deactive `chess` & `academic_governance` and ask another question with previous chat history, it does deny the request:
image
image

**(Such verbose error message are there because I completely disable the catch on `Exception` error in old `/cosmic` endpoint so that I know specifically what is the error about.)**

FaceWithDark:
Can I have both of you (@carlosclaiton & @Adnan525) to ask questions from old `/cosmic` endpoint instead? FE is mapped to call the new one (I believe so, after looking at the latest change on **this repo `Development` branch** & **FE `development` branch**), so we'd need to manually write the response on FastAPI instead. I have the example format here, feels free to edit the `user_message` field to trigger correct service:
```json
{
  "body": {
     "user": {
       "id": "019dec29-1c48-7e00-832c-cbb8c384b883",
       "role": "admin",
       "email": "u3295557@uni.canberra.edu.au"
     },
     "messages": []
   },
  "user_message": "How can you help me?"
}


// 1 History (this one trigger Academic Governance service, else error if disabled)
{
  "body": {
     "user": {
       "id": "019e05e0-413e-77f5-837f-8192b0e89882",
       "role": "admin",
       "email": "u3295557@uni.canberra.edu.au"
     },
     "messages": [
         {
	     "id": "019e05e0-413e-77f5-837f-8192b0e89882",
	     "role": "user",
	     "content": "How can you help me?"
	 },
         {
	     "id": "019e05b6-0d26-73ab-b7c5-1688c6bc207f",
	     "role": "assistant",
	     "content": "Hello! It seems like I'm equipped with a variety of capabilities to assist you. Here’s how I can help based on the services listed:\n\n1. **Chess Game Analysis**: If you need help predicting the next move in a chess game, just provide a sequence of moves or a FEN (Forsyth-Edwards Notation) string.\n2. **Vector Database Updates**: You can ask me to update the vector database with declarative sentences or PDF documents.\n3. **Code Assistance**: I can generate or improve code snippets and help answer questions related to coding.\n4. **General Knowledge and Reasoning**: I can answer general questions and provide reasoning based on available information.\n5. **Academic Governance**: If you have any queries related to academic governance, feel free to ask.\n\nCould you please let me know what specific assistance you need?"
	 }
     ]
   },
  "user_message": "What is Academic Governance?"
}
```

carlosclaiton:
I am not sure what you're asking about. All the tests we are doing should work with the old and new endpoints, as they should not change the content they send to the query analyser. The way we tested was by using FastAPI and older front-end versions that use the original endpoint.

FaceWithDark:
> I am not sure what you're asking about. All the tests we are doing should work with the old and new endpoints, as they should not change the content they send to the query analyser. The way we tested was by using FastAPI and older front-end versions that use the original endpoint.

Hmm... I tried this exact method in another [branch](https://github.com/TheOpenSI/CoSMIC/tree/COSMIC-120-new-approach) (this one is a sub-branch of [COSMIC-229](https://opensi.atlassian.net/browse/COSMIC-229)) and it does detect correct service per question asked (`Chess` & `Code Generation` service will definitely have error since for `Chess`, the stockfish binary file got removed in `Development` branch I think, and `Code Generation` need to enable the **PyCapsule** container but we didn't test this to work yet).

Would you mind tryout this branch first? If it does run for you and Adnan then I can merge it back to this PR and get things resolved.


At the high-level, the main change looks like this:
> **Before**
image

> **After**
image


The way to reproduce this bug flow:
```md
[Start CoSMIC container (5 services available by default)]
                           |
                           |
                           v
[Ask any question (assume it triggers system prompt)]
                           |
                           |
                           v
[CoSMIC responsed with 5 services along with its name and description (Correct)]
                           |
                           |
                           v
[Disable some services (let say 2)]
                           |
                           |
                           v
[Ask another question (same assumption)]
                           |
                           |
                           v
[CoSMIC still responsed with 5 services along with its name and description due to stall/cached
data. This happens because services data were fetched and called from `__init__` special method in 
`src/query_analyser/query_analyser.py`. This method will always be executed when the parent class
(`OpenSICoSMIC()`) get initiated from any other files (That's just how classes work). If we were to
fetch and call from `__call__` special method instead, this will only happen when the parent class
receive action from the CoSMIC system itself but not from Python (`__init__` will still executed
no matter how because of mentioned reason, but it won't create that "hallucinate" effect like what
we've indentified early on in this long convo)]
```

[COSMIC-229]: https://opensi.atlassian.net/browse/COSMIC-229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

FaceWithDark:
> [!IMPORTANT]
> Commit from c02551d is what I merged from the mentioned [branch](https://github.com/TheOpenSI/CoSMIC/tree/COSMIC-120-new-approach) above. However, I just realised that it actually a child of [COSMIC-129 branch](https://github.com/TheOpenSI/CoSMIC/tree/COSMIC-219) instead, hence why we see the git history of #124 here. I've put down notes for files/folders that belong to mentioned PR so that we don't consider those one as a part of the review in this PR.

backend/cores/dependencies.py
FaceWithDark, line 1: This's included as part of the merge from [sub-branch](https://github.com/TheOpenSI/CoSMIC/tree/COSMIC-120-new-approach), which is actually cloned from #124.

backend/cores/globals.py
FaceWithDark, line 1: This's included as part of the merge from [sub-branch](https://github.com/TheOpenSI/CoSMIC/tree/COSMIC-120-new-approach), which is actually cloned from #124.

backend/routers/cosmic.py
FaceWithDark, line 1: This's included as part of the merge from [sub-branch](https://github.com/TheOpenSI/CoSMIC/tree/COSMIC-120-new-approach), which is actually cloned from #124.

backend/routers/default_apis.py
FaceWithDark, line 1: This's included as part of the merge from [sub-branch](https://github.com/TheOpenSI/CoSMIC/tree/COSMIC-120-new-approach), which is actually cloned from #124.

backend/routers/models.py
FaceWithDark, line 1: This's included as part of the merge from [sub-branch](https://github.com/TheOpenSI/CoSMIC/tree/COSMIC-120-new-approach), which is actually cloned from #124.

docker/dockerfiles/fastapi.Dockerfile
FaceWithDark, line 1: This's included as part of the merge from [sub-branch](https://github.com/TheOpenSI/CoSMIC/tree/COSMIC-120-new-approach), which is actually cloned from #124.

src/services/llms/LLMBase.py
FaceWithDark, line 1: This will be deleted once I've confirmed test from @Adnan525 (or @carlosclaiton).

src/services/llms/Ollama.py
FaceWithDark, line 1: This's included as part of the merge from [sub-branch](https://github.com/TheOpenSI/CoSMIC/tree/COSMIC-120-new-approach), which is actually cloned from #124.

api.py
FaceWithDark, line 1: This's included as part of the merge from [sub-branch](https://github.com/TheOpenSI/CoSMIC/tree/COSMIC-120-new-approach), which is actually cloned from #124.

compose.yaml
FaceWithDark, line 1: This's included as part of the merge from [sub-branch](https://github.com/TheOpenSI/CoSMIC/tree/COSMIC-120-new-approach), which is actually cloned from #124.

file_list.csv
FaceWithDark, line 1: This will be deleted once I've confirmed test from @Adnan525 (or @carlosclaiton).

scripts/configs/config.yaml
FaceWithDark: Reverted.

FaceWithDark:
@carlosclaiton @Adnan525 if the convo on this PR gets too long and complex (mainly due to issue keep coming after each review), would any of you mind if I closed this PR and break the changes into smaller JIRA tickets with smaller PR as well? This one doesn't do much changes but because of the amout of image linked in here polluted the chat a lot and it's hard to scroll to correct part as well. The latest commit I did does make **CoSMIC** & **Query Analyser** running like how it normally be in `main` branch.

carlosclaiton:
> @carlosclaiton @Adnan525 if the convo on this PR gets too long and complex (mainly due to issue keep coming after each review), would any of you mind if I closed this PR and break the changes into smaller JIRA tickets with smaller PR as well? This one doesn't do much changes but because of the amout of image linked in here polluted the chat a lot and it's hard to scroll to correct part as well. The latest commit I did does make **CoSMIC** & **Query Analyser** running like how it normally be in `main` branch.

As I mentioned on the table, I think we should close those PR and start again.
It should require modify at maximum only three python files:
qa.py, query_analyser.py and OpenSI_Cosmic.py (the last one if you want to do the imports here).

I have discussed it with @Adnan525, if you want to have a second go I am happy. But I suggest start again fresh from Development and modify as minimum as possible to facilitate the review.

Adnan525:
i agree with Carlos, this would probably work fine now but merging this without understanding the codebase properly would be a nightmare for maintenance. i will close both of the PRs (#117 and #124) by EOD unless you choose to do so earlier. i will look into the requirements, hopefully next week and see what we can do to bring those changes in. thanks

FaceWithDark:
All goods as long I received the confirmation from both of you so that I can starting breaking the changes into smaller part. Hopefully by this time, it's much easier to understand the modification.

carlosclaiton:
@FaceWithDark , I am closing this PR as mentioned above, you can submit another one as per our conversation if you want to have a second go with this. 

We should focus in keeping the PR short and small modifications to ensure code logic is kept the same.

PR #118, opened 2026-04-21
Cosmic 188: Add Qdrant vector store integration

marslanshaukat:
Replaced existing vector storage with Qdrant


carlosclaiton:
This implementation have also modified CoSMIC logic, it incorporates the document into the response about the system:

Screenshot 2026-04-22 at 11 54 43 am

marslanshaukat:
The issue is due to QA not able to understand the intent and router calling the wrong service. Need to investigate further for best possible solution. Should we use the same PR for enhancement or create a new one?

carlosclaiton:
> The issue is due to QA not able to understand the intent and router calling the wrong service. Need to investigate further for best possible solution. Should we use the same PR for enhancement or create a new one?

I don't think that is the issue, this part is handled in a bit differently. We use the same PR, just commit the solutions, and when we manage to get the expected behaviour, we merge.

marslanshaukat:
Issue resolved.
Known issues include a delay in the first response or getting stuck in that case, try refreshing the frontend.

src/query_analyser/query_analyser.py
```diff
@@ -359,11 +382,14 @@ def __call__(
...
             "system_information_relevance": False,
-            "system_information": ""
+            "system_information": "",
+            "needs_context": False # for Routing
         }
 
         if self.service_index >= 0:
             service_option = str(self.service_index)
+            # Default needs_context to True for manual service selection if it's Academic QA
+            service_info_dict["needs_context"] = True if service_option == "4" else False
```
carlosclaiton: Why service 4 has a special treatment?

src/query_analyser/user_prompt.py
```diff
@@ -58,13 +58,24 @@ def __call__(
...
+
         user_prompt: str = "{0:s}, {1:s}. {2:s}; {3:s}. {4:s}. {5:s}.".format(
             f"Given {self.num_services} services: '{self.service_string}'",
-            f"which service can answer the following query? The query is '{question}'",
-            "For instance, if the query is to predict the next chess move, then select service 0",
-            "otherwise, if the query is to generate or modify a code, then select service 2",
-            "otherwise, if the query is about Academic Governance, then select service 4",
-            "Just return which service without any explainations"
+            f"determine which service can answer the query: '{question}'",
+            "Also decide if external context (RAG) is required to answer correctly",
```
carlosclaiton: Where is the information about the chess service? why service 4 would here required special treatment again?

carlosclaiton (changes requested):
I am not sure why do you require to modify CoSMIC logic to add a new vector database

src/query_analyser/query_analyser.py
marslanshaukat: Before the logic for General QA (the else block) executes retrieval immediately if is_rag is set to true and that is why QABASE was treating both General QA and Academic QA identically. We can revert it to always check RAG get context -> check it using QA and then select other service or we can keep this change where QA analyse the prompt select the service based on it.

src/query_analyser/user_prompt.py
marslanshaukat: We can also add chess service in example but as its straight forward call the LLM should understand that intent. We can test it further.

src/query_analyser/query_analyser.py
carlosclaiton: QA has to treat General Question and Academic QA the same as respect of RAG. General question can have RAG capability. System questions are the one that does not requires RAG capability. Previous logic of qa was working fine, when I added the Academic service it was working fine. We don't want a solution that everytime we add a new service we need to change the qa.

carlosclaiton (changes requested):
The logic is still not working as originally, we don't want to have to change query_analyser.py every time we add a new service. It was working fine before you implemented the new vector database. 

Maybe we do need to have a boolean variable in the services table for the database that informs when a service requires RAG capability?

marslanshaukat:
Made changes without changing the overall logic. Please review and let me know. Thanks.

src/services/qa.py
```diff
@@ -186,7 +186,10 @@ def __call__(
             raw_response, response = self.code_generator(query)
 
         else:
-            if is_rag:# General question has RAG activated
+            RAG_ENABLED_SERVICES = ["4"] # Academic QA triggers retrieval
```
carlosclaiton, line 189: Is this solution will also give RAG capability to general questions?

carlosclaiton: Added a todo into the source code to double check this solution

carlosclaiton (approved):

PR #120, opened 2026-04-22
bug: open the port to allow us to use ollama api and also detect that…

carlosclaiton:
qwen3.5 makes the system very slow.

Change the default model to qwen2.5 in the config
Modify Docker Compose to expose Ollama port


Adnan525 (approved):
tested with correct sequence, working and response time is good.

PR #121, opened 2026-05-02
add: integrated cosmic-V1 with bings' POST and PATCH apis to add data…

manilesrun:
Hi prof, 

this is the cosmic-api integrated with bing's apis POST and PATCH. Could you please review?


carlosclaiton (approved):
it is a good ideas to have version control to the API

PR #124, opened 2026-05-11
Fixing double initiallisation every hot-reload during development process in CoSMIC [COSMIC-229] (Old)

FaceWithDark:
**Re: [COSMIC-229 Ticket](https://opensi.atlassian.net/browse/COSMIC-229?atlOrigin=eyJpIjoiNWZjYmRhMzZjNWRkNDI2NGE2MjViNmVlNzY1NGY1YmEiLCJwIjoiaiJ9)**

Extra update along ticket description information:
> **Before**
image

> **After**
image

> [!NOTE]
> **Due to several new changes on my other PRs (#117), this one would need to get reviewed and merged first so that the git history doesn't looks oddly weird or duplicated.**

PR #127, opened 2026-05-14
Update `CoSMIC` README to match `COSMIC-DB` structure [COSMIC-219]

FaceWithDark:
**Re: [COSMIC-219 Ticket](https://opensi.atlassian.net/browse/COSMIC-219?atlOrigin=eyJpIjoiMmI0OTg3YWYwMTY5NDJhYmJlYWFhMmY3YWMwZGM4NWMiLCJwIjoiaiJ9)**

Extra update along ticket description information:

I'm proposing some new folders here in the **`Project Hierarchy`** section that aren't yet implemented in CoSMIC (or maybe never). I want to see your opinions on those if the repo structure should follow this style or not. In details:
- Merge files from `/default` folder --> `/data` folder
- Merge files from `/modules` folder --> `/src/services/<service name>` folder depending on the purpose of each current files
- Remove `/pipelines` folder since we don't rely on [OpenWebUI](https://github.com/open-webui/open-webui) anymore
- Remove `/scripts` folder. Files that look like for testing purposes can go to `/tests` folder
- Move `/src/services/llms` folder to `/src/models` folder with explanation in the proposed repo structure
- Re-structure the reset of `/src` folder according to proposed repo structure
- Re-name `/third_party` folder --> `/bins` folder for more general purpose
- Move `api.py` file to `/backend/cores` folder to be consistent with repo structure in [COSMIC-DB repo](https://github.com/TheOpenSI/COSMIC-DB/tree/dev)
- Rename `docker-compose.yaml` file --> `compose.yaml` file for the same purpose (change already done in #124)
- Move `demo.py` file to `/examples` folder
- Modify `main.py` file a little bit to act more like an entry point file *(TODO: will have another commit when proposed repo structure agreed)*
- Move `OAuth.md` file to `/docs` folder
- Move `start.sh` file to `/bins` folder
- Move `test.ipynb` file to `/tests` (or `/examples`) folder
- Temporary move `to_be_removed.txt` file to `/docs` folder until we get to the stage that need to touch on it
- The rest of new folders are well explained in the proposed repo structure

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,50 +14,36 @@
 CoSMIC/
 ├── backend/
 │   ├── cores/                    # Core FastAPI application setup
-│   └── routers/
-│       ├── api_endpoints/        # CoSMIC-only API endpoints (e.g, pulling Ollama models from `/models` endpoint)
-│       └── normal_endpoints/     # CoSMIC non-API endpoints (e.g, sending user queries to Query Analyser using `/cosmic`)
-├── bins/                         # 3rd party binaries and vendored tools (e.g. Stockfish engine, FEN board renderer)
+│   └── routers/                  # CoSMIC API & non-API endpoints
 ├── data/                         # Datasets in certain format (e.g., CSVs, Excels, etc)
+├── default/                      # Dataset templates (TODO: to be merged into `/data` directory)
 ├── docker/                       # Containerisation resources and orchestration files
-│   ├── configs/                  # Non-sensitive config files for Docker services
 │   ├── dockerfiles/              # Dockerfile for each service defined in the Compose file
-│   └── secrets/                  # Secure storage for sensitive data (e.g., model special credentials)
-├── docs/                         # Project documentation (e.g., OAuth2 guide)
-├── assets/                       # Static files (e.g., favicon)
-├── examples/                     # Standalone runnable examples and demo scripts
-├── prompts/                      # Prompting logic for Query Analyser, specialised services, or SLM abstractions
-│   ├── system_prompts/           # Define the knowledge base and behavioural boundaries of each middleman layer
-│   │   ├── query_analyser/       # Instructs Query Analyser on how to classify and route incoming queries
-│   │   ├── services/             # Instructs each specialised service on its domain knowledge and constraints
-│   │   └── slms/                 # Instructs the SLM abstraction layer on output format and interaction rules
-│   └── user_prompts/             # Define how each middleman layer re-forms the end-user's input before forwarding it
-│       ├── query_analyser/       # Structures the raw query into a well-formed classification request for the SLM
-│       ├── services/             # Restructures the query into a precise, domain-optimised prompt for the target SLM
-│       └── slms/                 # Applies any final formatting or chain-of-thought (CoT) scaffolding before SLM inference
-├── src/
-│   ├── models/                   # LLM class definitions and abstractions (e.g., Ollama, ChatGPT, Mistral)
+├── modules/                      # Subservices for each of CoSMIC services (if any) (TODO: to be re-structured inside `/src` directory)
+├── pipelines/                    # Depricated CoSMIC pipeline logic to work with OpenWebUI (TODO: to be removed from new structure)
+├── scripts/                      # Standalone runnable examples and demo scripts
+│   ├── configs/                  # Default config data for Query Analyser to use depends on selected service (TODO: to be removed and call from BE instead)
+│   └── <others>/                 # Unit test files (TODO: to be moved into its own `/tests` folder on project root)
+├── src/                          # This is where the "heart" of CoSMIC located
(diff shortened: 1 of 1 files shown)


Adnan525:
Please add COSMIC-DB setup info

Adnan525:
Please fix the requirements for Native and Docker usage, remove Zohaib from the contact list

FaceWithDark:
> Please add COSMIC-DB setup info

Done as of commit 755c989

> Please fix the requirements for Native and Docker usage, remove Zohaib from the contact list

Done as of commit 188269a & 0d300c1

Force pushed due to update commit message for correct commit category.

Adnan525 (changes requested):
- For Native setup, ollama is only optional if user is using OpeanAI keys and please remove the icons
- why "sudo" for docker in linux or macos, user should be added to the docker group which is more common than users having sudo access.
- Please remove "Therefore, feels free to go and grab a coffee, say good morning to your AI agents, get the kids to school, etc while waiting and CoSMIC will be ready when you are. Or you sit there and monitor the progress with this command (I would suggest not to since we all know why):"
- "Before you begin, ensure that you have Python (v3.14+), Ollama (Optional) and uv running on your system:" you cant say ensure these are running and also make ollama optional. it's optional only when we use openai key
- same issue - "Start a Local SLM via Ollama (Optional)" i think you are mixing up gpu with ollama. as far as i know if we dont use OpenAI key, we must use ollama since we havent used the HF implementation for a while and ollama is the default setup at the moment

FaceWithDark:
> * For Native setup, ollama is only optional if user is using OpeanAI keys and please remove the icons
> * why "sudo" for docker in linux or macos, user should be added to the docker group which is more common than users having sudo access.
> * Please remove "Therefore, feels free to go and grab a coffee, say good morning to your AI agents, get the kids to school, etc while waiting and CoSMIC will be ready when you are. Or you sit there and monitor the progress with this command (I would suggest not to since we all know why):"
> * "Before you begin, ensure that you have Python (v3.14+), Ollama (Optional) and uv running on your system:" you cant say ensure these are running and also make ollama optional. it's optional only when we use openai key
> * same issue - "Start a Local SLM via Ollama (Optional)" i think you are mixing up gpu with ollama. as far as i know if we dont use OpenAI key, we must use ollama since we havent used the HF implementation for a while and ollama is the default setup at the moment

I've updated to mark Ollama as mandatory installation on **Native setup** instead so it isn't much of an issue whether API key is being used or not.

I've already mentioned the reason on why using `sudo` to run any **Docker commands** on Linux/MacOS within the README, so I don't see any problems here.

The rest of your review I've applied in commit [1dff0bf](https://github.com/TheOpenSI/CoSMIC/commit/1dff0bf3c36c2a9d790b2ca0d9b443e67d4fd0d8).

Adnan525:
Not quite, mate. You've mentioned that sudo is needed and pointed to rootless mode as an alternative, but you haven't actually explained why. There's a big difference between the two.

think about it from a fresh pair of eyes, someone who's never used Docker before. The very first thing they're going to hit is a permission error when they try to run a docker command without sudo. At that point they'll either blindly copy-paste the sudo version, or go to their IT admin and ask for access. And the first thing any admin worth their salt will ask is "why does this application need sudo on my machine?" - and right now the README gives them absolutely nothing to work with there.
The real answer is that Docker's daemon runs as root because it needs deep access to the host system. things like kernel namespaces, cgroups, and the network stack. By default, only root or users in the docker group can talk to the daemon socket at /var/run/docker.sock. So the proper fix isn't actually sudo at all, it's adding the user to the docker group, which is the standard recommended approach. Sudo is more of a quick workaround, and not a particularly great one to lead with in documentation because it's not how most teams operate day-to-day.
The rootless mode note is also a bit buried it shows up after the command, with no context explaining what problem it solves or who it's even relevant for. Someone who doesn't already know what rootless Docker is will just skip right past it.
So it's not that the README is wrong, it's just incomplete in a way that will cause real friction for the people who need it most. hope this makes sense

FaceWithDark:
I get what you meant when I default to use `sudo` on pretty much all **Docker commands** being used in the README. The thing is that on Linux, vast majority of distros (including popular one like **Ubuntu or Fedora**) out there doesn't auto setup rootless Docker when it get installed, but rather the user (IT admin/department in this case) have to set it up themself through the selected distro's docs page or directly from [Docker Docs](https://docs.docker.com/engine/security/rootless/) page.
I've no issues to simply assume that user have a rootless Docker setup and run the provided command as it is without problems. But again, this would very much be a Linux-only matter as if I'm not wrong on MacOS, it works similar to Windows (just recently) as both of them are running through a small VM under the hood with rootless enable by default anyways.

FaceWithDark:
**UPDATE:**
- `sudo` removed from each Docker command in the `README.md` file (commit f1ec6d1)

carlosclaiton (approved):
Thank you, looks good

PR #129, opened 2026-05-21
add: upload files into the chat_session_memory

manilesrun:
Hi @marslanshaukat, 

this is the simple upload file into the chat_session_memory api, just upload file, then the api will create path, read the file and then write the content into the folder based on the path. 

- we still need to further develop it like file size limit, file type, ... ; however, i think it is basic enough for you to continue the task cosmic-227 

- i have checked the docker container already, the file uploaded is in there

### DIFF
--- a/api.py
+++ b/api.py
@@ -7,7 +7,7 @@ from fastapi.middleware.cors import CORSMiddleware
 
 
 ### Internal modules ###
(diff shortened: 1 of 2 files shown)


marslanshaukat:
good work. It seems its not integrated with the front end yet. But safe to merge with cosmic back-end.

PR #130, opened 2026-05-24
Update CoSMIC's Docker Compose file naming from `docker-compose.yaml` to `compose.yaml` for consistent convention across modular repos [COSMIC-245]

FaceWithDark:
*Ref: [COSMIC-245 ticket](https://opensi.atlassian.net/browse/COSMIC-245?atlOrigin=eyJpIjoiNThhZGU5YjNkNjIxNDZjNDgyNDc1NDRlYTc3NGRjNGQiLCJwIjoiaiJ9)*

**Extra information besides what have been explained in ticket:**
- N/A

PR #132, opened 2026-05-24
Fix double initiallisation issue during development on `CoSMIC` repo [COSMIC-229] (New)

### LINKED ISSUE #125: Double initialisation when starting and developing `CoSMIC` repo
> [!IMPORTANT]
> Issue tracker is **ONLY** used for reporting bugs. New features should be discussed during our internal meetings. Please use [stackoverflow](https://stackoverflow.com), etc for supporting issues.

<p align="center"><i>(notice the duplication not here)</i></p>

<p align="center"><i>(notice the double initialisation)</i></p>

1. Remove previously built Docker services by executing `docker compose down --volumes --rmi all`. 
2. Fresh built all Docker services again by executing `docker compose up --build -d`.
3. After all services are up and Ollama model installed, look at messages after those and notice similar pattern repeat twice before the line `Application startup complete` show.

> During development:
1. Do any changes to any **Python files** and hit save.
2. FastAPI hot-reload feature activated and notice similar pattern repeat twice before the line `Application startup complete` show.

## Detailed Description
My machine is not capable of running the deafault Ollama model so I have to downgrade and use `llama3.2:1b` version. However, even with this smaller model, it would still take around 10 - 15s to see new changes apply.

Therefore, I think the issue is lying somewhere within the codebase that 'cause this accidental double trigger effects.

FaceWithDark:
*Ref: [COSMIC-229 ticket](https://opensi.atlassian.net/browse/COSMIC-229?atlOrigin=eyJpIjoiNmFhNWRmNTE5NmY2NDg5YmIwNTc5OGEzMDVkM2NjMDIiLCJwIjoiaiJ9)*

> [!IMPORTANT]
> Make sure to merge #130 first due to the change to **Docker Compose** file name. This PR would probably have merge conflict when that PR get merged but it's an easy fix.

- *Before bugfix:*
> image

- *After bugfix:*
> image

- _97792b7_ - Change **FastAPI** entrypoint var name from `app` → `cosmic_app` (matching convention in `CoSMIC_DB` repo)
- _c1c1ef4_ - Update `backend/routers/default_api.py`’s endpoints to adapt with the bugfix solution
- _466da19_ - Update `backend/routers/cosmic.py`’s endpoints to adapt with the bugfix solution
- _c1c1ef4_ - Create dedicated `backend/cores/dependencies.py` file to manage all **FastAPI’s `Depends()` classes**, which being used in the proposed bugfix solution


FaceWithDark:
**Forced pushed to resolve conflicts on this PR when #130 got merged into `Development` branch.**

api.py
Adnan525, line 1: - please keep the debug py code at the very top, at least before the app state initilisation. 
- the comment "        # No API key provided? No problem at all, we'll just use local-first SLMs
        # like Ollama instead" in the else block is confusing. That block would trigger when there is an env key.
-  please dont try to fix multiple things in the same PR. that app -> cosmic_app could be done in a separate PR but it's okay for now

docker/dockerfiles/fastapi.Dockerfile
Adnan525, line 1: - does this have anything to do with the double init issue?
- not sure why are we using aliases - base, setup, uv?

FaceWithDark: I was trying to fix 2 things on the same PR, I think. This doesn't affect the double init issue but more like a DevOps thing.

For the aliases, if I'm not wrong, that's one of the recommended practices when creating Dockerfile for multiple environments. I can't find the actual source I read about this but the idea of using aliases was inspired from the [example Bun Dockerfile setup here](https://bun.com/docs/guides/ecosystem/docker).

backend/routers/default_apis.py
Adnan525, line 1: I can see you removed the old cosmic init code using globals and using dependency injection pattern. this is great but you need to understand that from a reviewer's point of view this is extremely hard to keep track of and I believe @carlosclaiton  explicitly mentioned to focus on one isssue at a time

api.py
FaceWithDark: Yeah, I just realise I was trying to fix 2 things at the same time. Sorry for that.

About the comment, I'll fix it up on the new commit. I understand this comment
```python
# No API key provided? No problem at all, we'll just use local-first SLMs
# like Ollama instead
```
is quite misleading but essentially, I was just trying to make the comment more engaging to read. This should goes to the `if` case instead as if no API key provided, we can assume that we using models pulled from [Ollama](https://ollama.com/models). I was trying to write that `if/else` block in a **defensive style** *(invalid case first, valid case second)* though I think I forget to double-check it.

backend/routers/cosmic.py
Adnan525, line 1: on this we have refactoring imports, restructuring the pydantic models, renaming the docker service url, fixing a user_role payload bug anf of couse removing globals. bundled together they make the diff harder to review and harder to roll back if something goes wrong.

on the rebuild logic, this block that checks whether the config timestamp or API key has changed and reconstructs OpenSICoSMIC is now duplicated between cosmic.py and default_apis.py. Since we already have dependencies.py for exactly this kind of shared logic, could you extract it into its own function there? Something like rebuild_if_changed(request: Request) that both endpoints can pull in via dpends(). That way if the rebuild logic ever needs to change, there's only one place to update it.

Adnan525 (changes requested):
please refer to individual comments

docker/dockerfiles/fastapi.Dockerfile
Adnan525: thats a good reference and i can see how this can be useful in a proper mult stage build, but since this dockerfile only has a single stage doing all the work, the aliases don't actually do anything here

backend/routers/cosmic.py
FaceWithDark: I could do that but then I'm thinking, since we're going to mainly use the new `/cosmic` endpoint from now on and complete replace/delete the old one, having a share func (or similar) would create more complexity. I still apply the changes in both endpoints here mainly because @carlosclaiton would test this from FE, rather than directly from the `/docs` page *(this's just my assumption)*. I'm having a hard time to test the new one easily without having to turn on FE, hence why I just do all of my tests through the old endpoint instead.

I do planned to have a look over this new endpoint that @manilesrun built as I think there must be a nicer way (for me at least) to test this endpoint purely from the 2 `/docs` pages. However, I need to wait for this PR to get approved and merged first or else, there'll likely be another **1k+ code** changes occur.

FaceWithDark: And yes, I should have break these changes into at least 2 - 3 smaller commits, especially when some of the new updates are just for the code style (e.g., different import styling). Again, I'm apologise about this. I did apply what I said here on #134 as a way of me reflected myself for this kind of matters.

backend/routers/default_apis.py
FaceWithDark: I'll take a **$\color{red}{\text{big red note}}$** on this and make sure to break the changes smaller (either more commit or multiple small PRs). My reasoning are there within the comment in `backend/routers/cosmic.py` file so I won't repeat it again here.

docker/dockerfiles/fastapi.Dockerfile
FaceWithDark: If you take a closer look on the log during `docker compose up --build -d`, You'd see my setup for multi stage build does get invoked. It's not benefit for now since we aren't doing anything too complex at the moment. However, I do think this's helpful as we get closer to the new production vesrion.

api.py
FaceWithDark: Updated as of commit c6f0f7d & cc55b9c

FaceWithDark:
> please refer to individual comments

Thank you for the reviews, @Adnan525. I've replied to all of them and resolved those that I think it's suitable without further explanations.

**For the force pushed, it's due to re-wording my pushed git commit message. Since those are technically new *commits*, I've to overwrite its git history.**

Adnan525:
hi, please fix the conflicts and it's good to merge. I tested with the `/cosmic` endpoint

FaceWithDark:
> hi, please fix the conflicts and it's good to merge. I tested with the `/cosmic` endpoint

Hi Adnan, I've resolved the conflicts and test running this branch again to make sure it correct (it was done purely on the PR itself, rather than having to do it in the terminal).

FaceWithDark:
Afternoon @Adnan525, I've just resolved the new conflicts since #138 got merged to `Development` branch. Could you do a final run and get this to merged status?

Adnan525:
All tested

carlosclaiton (approved):
After discussion with Adnan, I approved to be merged

carlosclaiton (commented):
File "/usr/local/lib/python3.14/contextlib.py", line 214, in __aenter__
    return await anext(self.gen)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/app/api.py", line 107, in lifespan
    opensi_cosmic_instance: OpenSICoSMIC = OpenSICoSMIC(config_path=str(NEW_CONFIG_PATH))
                                           ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/src/opensi_cosmic.py", line 65, in __init__
    self.config_path: Path = Path(config_path).resolve(strict=True)
                             ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/usr/local/lib/python3.14/pathlib/__init__.py", line 938, in resolve
    return self.with_segments(os.path.realpath(self, strict=strict))
                              ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "<frozen posixpath>", line 457, in realpath
FileNotFoundError: [Errno 2] No such file or directory: '/app/scripts/configs/config_updated.yaml'

FaceWithDark:
> File "/usr/local/lib/python3.14/contextlib.py", line 214, in **aenter** return await anext(self.gen) ^^^^^^^^^^^^^^^^^^^^^ File "/app/api.py", line 107, in lifespan opensi_cosmic_instance: OpenSICoSMIC = OpenSICoSMIC(config_path=str(NEW_CONFIG_PATH)) ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/app/src/opensi_cosmic.py", line 65, in **init** self.config_path: Path = Path(config_path).resolve(strict=True) ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^ File "/usr/local/lib/python3.14/pathlib/**init**.py", line 938, in resolve return self.with_segments(os.path.realpath(self, strict=strict)) ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^ File "", line 457, in realpath FileNotFoundError: [Errno 2] No such file or directory: '/app/scripts/configs/config_updated.yaml'

@carlosclaiton , it's from this line:
```python
OG_CONFIG_PATH:     Path = Path(__file__).resolve(strict=True).parent.joinpath(
    "scripts",
    "configs",
    "config.yaml"
)
NEW_CONFIG_PATH:    Path = OG_CONFIG_PATH.parent.joinpath("config_updated.yaml")
```

Only happens on fresh run since the file is expected to be found beforehand. You can run this to temporary fix it:
```bash
docker cp ./scripts/configs/config.yaml cosmic-fastapi:/app/scripts/configs/config_updated.yaml
```

I didn't try to get this correct since we're going to stop relying on this YAML file as soon as we finish migrating the `/config` endpoint.

carlosclaiton:
> > File "/usr/local/lib/python3.14/contextlib.py", line 214, in **aenter** return await anext(self.gen) ^^^^^^^^^^^^^^^^^^^^^ File "/app/api.py", line 107, in lifespan opensi_cosmic_instance: OpenSICoSMIC = OpenSICoSMIC(config_path=str(NEW_CONFIG_PATH)) ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/app/src/opensi_cosmic.py", line 65, in **init** self.config_path: Path = Path(config_path).resolve(strict=True) ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^ File "/usr/local/lib/python3.14/pathlib/**init**.py", line 938, in resolve return self.with_segments(os.path.realpath(self, strict=strict)) ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^ File "", line 457, in realpath FileNotFoundError: [Errno 2] No such file or directory: '/app/scripts/configs/config_updated.yaml'
> 
> @carlosclaiton , it's from this line:
> 
> ```python
> OG_CONFIG_PATH:     Path = Path(__file__).resolve(strict=True).parent.joinpath(
>     "scripts",
>     "configs",
>     "config.yaml"
> )
> NEW_CONFIG_PATH:    Path = OG_CONFIG_PATH.parent.joinpath("config_updated.yaml")
> ```
> 
> Only happens on fresh run since the file is expected to be found beforehand. You can run this to temporary fix it:
> 
> ```shell
> docker cp ./scripts/configs/config.yaml cosmic-fastapi:/app/scripts/configs/config_updated.yaml
> ```
> 
> I didn't try to get this correct since we're going to stop relying on this YAML file as soon as we finish migrating the `/config` endpoint.

The temporary solution does not work for being in development branch, code there has to be fully functional, not complete but functional. 

Maybe we put a pause in this one, and get the reading from the DB first, it should only have to modify the file called opensi_cosmic.py

PR #134, opened 2026-05-25
CoSMIC read services from DB [COSMIC-120] (New)

FaceWithDark:
*Ref: [COSMIC-120 ticket](https://opensi.atlassian.net/browse/COSMIC-120?atlOrigin=eyJpIjoiMWY2YjkxZTllZjkyNDRmMzk2YzUwZGQzNjcwNDA1MzEiLCJwIjoiaiJ9)*

> [!IMPORTANT]
> Make sure to merge #130 first due to the change to **Docker Compose** file name. This PR would probably have merge conflict when that PR get merged but it's an easy fix.

- _54acfce_ - linter updates to `src/opensi_cosmic.py` file
- _da91eb5_ - linter updates to `src/services/qa.py` file
- _e164989_ - linter updates to `src/query_analyser/query_analyser.py` file
- _2c17de7_ - stale data issue found when testing to get CoSMIC system info
- _28256e8_ - add new method to dynamic grab **service name & desc** to prepare for next couple commits that pass those data to `QABase.__call__()` & `QueryAnalyser.__call__()` dunder method accordingly.
- _e2a2e26_ & _7d8315c_ - adjust the hard-coded services name in `QABase.__call__()` & `QueryAnalyser.__call__()` dunder method accordingly to the new grabbing method.
- _9780b05_ & _66fee2a_ - add logic to handle empty active services fetched from our API endpoint.


FaceWithDark:
> [!NOTE]
> I leave debug messages and debug mode for `OpenSICoSMIC.get_services()` method enabled so that when review/test, it's easier to track down what's happen behind the scene. I'm not yet used to using a debugger so I'm apologise in advance if it's much harder to track down issues within the **Docker container log**.

> [!IMPORTANT]
> I moved `QueryAnalyser.__call__()` dunder method next to `QueryAnalyser.__init__()` dunder method to make it match the style in `qa.py` file. I do believe this would help tracking down future issues easier since we mostly call all the classes as if it's a function call, hence that `__call__()` dunder method occurs.

FaceWithDark:
> [!WARNING]
> Most of the time when I ask a question to CoSMIC, it does responed back with LLM responses for the right service (double-check with debug messages) **when all services available** AND **when one specific service available**. However, when there're 2 or slightly more, CoSMIC somehow doesn't work right and having issue with the *auto mode (-1)*. This might be further logic issues that aren't related to the scope of this PR, or something else that I couldn't think off while writing these lines.

**PR is ready for review now. Debug and log messages will get deleted as soon as the PR is accepted for a merge.**

src/query_analyser/query_analyser.py
```diff
@@ -94,6 +78,85 @@ def __init__(
...
+
+        Returns:
+            service_option (str):
+                service option.
+
+            service_info_dict (dict):
+                updated information dictionary.
+        """
+        # Set a list of services.
+        # TODO:
```
carlosclaiton, line 115: What is the TODO, is not that what the function is doing already?

src/query_analyser/query_analyser.py
```diff
@@ -152,13 +327,14 @@ def mapping(
                         information=f"Unknown service '{option}' from '{response}'."
                     )
                 )
-
                 return "-1"
+
+            else:
+                return option
+
```
carlosclaiton, line 334: why you change the order onf the else? It seems the final response will not be the option anymore.

src/query_analyser/query_analyser.py
carlosclaiton, line 1: During the build we have a following warnings:
/app/src/query_analyser/query_analyser.py:315: SyntaxWarning: "\d" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\d"? A raw string is also an option.
  pattern="service (\d{1,3}\.\d{1,3}|\d{1,3})",
/app/src/query_analyser/query_analyser.py:380: SyntaxWarning: "\[" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\["? A raw string is also an option.
  pattern="[\[,\:](.*?[,\s].*?)[\.,\]]?$",
/app/src/query_analyser/query_analyser.py:388: SyntaxWarning: "\/" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\/"? A raw string is also an option.
  "(((?:[rnbqkpRNBQKP1-8]+\/){7})[rnbqkpRNBQKP1-8]+)",
/app/src/query_analyser/query_analyser.py:389: SyntaxWarning: "\s" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\s"? A raw string is also an option.
  "\s([b|w])\s(-|[K|Q|k|q]{1,4})\s(-|[a-h][1-8])\s(\d+\s\d+)$"
/app/src/query_analyser/query_analyser.py:451: SyntaxWarning: "\:" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\:"? A raw string is also an option.
  pattern="(?<=\:\s)(.*?)+\.pdf",
/app/src/query_analyser/query_analyser.py:478: SyntaxWarning: "\:" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\:"? A raw string is also an option.
  pattern="\:((\"|\')?(.*?)[\",\']?$)",

carlosclaiton, line 1: something is not quite right. 

I have noticed that if we remove only chess from the list it crashes the system

I used the Development branch and commented out the Chess service from the dictionary to check the behaviour of the system before this change:

        # Set a list of services.
        self.services = {
            # "0": "if it is a chess game, predict the next chess move by providing a sequence of moves or a FEN",
            "1": "update the vector database with a declarative sentence (not a question) or a pdf document",
            "2": "generate or improve a code or answer a question in order to generate or improve a code",
            "3": "answer a question or provide a reasoning, which cannot be achieved by the other services",
            "4": "Answer question about Academic Governance"
        }

This issue still there, this is some previous problem with the logic of the query_analyser, It seems it requires always a service with index 0 (It would be nice to track this issue before we move on with this PR.

I have also noticed that it works fine if you have chess capabilities with more than one service or only one service on the list. 
If we have memory capability with only two more services that tend always to selected the memory service. 

We do need to use some break point to check what is causing the return the wrong option and if that is because of the changes on the __call__

carlosclaiton (changes requested):
Please have a look at my comments

src/query_analyser/query_analyser.py
FaceWithDark: I forget to delete this when breaking changes into smaller, trackable git commits.

src/query_analyser/query_analyser.py
FaceWithDark: Previously, this method would always return `-1 (auto mode)` no matter if the truncated response have the service index or not. From here I think: why don't we just return the selected index directly if SLM response does *correctly (I can't gurantee about this since user question can be tricky to guess)* pick out a service so we can further improvise it and sent the final answer back to user after re-prompting technique?

src/query_analyser/query_analyser.py
FaceWithDark: These're just warnings about using regex pattern as literal string in `search()` function. It doesn't affect the regex logic or anything since **Python** still gladly accept and treat it as if we providing regex pattern in the *right way*. It was like this from the `main` branch's codebase long time ago. I can update it to use raw string (they called it `r-string` in **Python**) but I want to make sure the core changes to work first.

FaceWithDark: That's what I've been experienced while testing to toggle on/off some services to make sure my dynamic changes work just like the hard-coded version. I did mention about this on the general PR comment. Very likely we need to create another PR to track down this issue, then merge back to this PR to confirmed the dynamic updates.

carlosclaiton (approved):
I have approved as the bug was not introduced by this PR, then we can fix this another time.

PR #137, opened 2026-06-02
CoSMIC read services from DB [COSMIC-120] (Continue)

FaceWithDark:
*Ref: [COSMIC-120 ticket](https://opensi.atlassian.net/browse/COSMIC-120?atlOrigin=eyJpIjoiMTQzOTM3NWIxMTExNDY5ZDgyZTA4NGJjYmE3Y2YxMjEiLCJwIjoiaiJ9)*

> [!NOTE]
> This PR is here just to address my comments in #134 about enable all debug/log messages by default so that reviewers can see the same testing result like I do.

### DIFF
--- a/src/opensi_cosmic.py
+++ b/src/opensi_cosmic.py
@@ -275,7 +275,7 @@ class OpenSICoSMIC:
                 retrieve_score
             ) = self.qa(
                 query=question,
-                services=self.get_services(verbose=True),
+                services=self.get_services(),
                 # The context that we are passing here is chat history. See api.py
                 context=context,
                 is_rag=True,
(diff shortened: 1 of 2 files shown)


carlosclaiton (approved):
Great work, you can merge it.

src/opensi_cosmic.py
```diff
@@ -275,7 +275,7 @@ def __call__(
                 retrieve_score
             ) = self.qa(
                 query=question,
-                services=self.get_services(verbose=True),
+                services=self.get_services(),
```
carlosclaiton, line 278: is False the default?

FaceWithDark: yes

PR #142, opened 2026-06-04
chore: comment out the debug messages when Query Analyser does found active services

FaceWithDark:


carlosclaiton (approved):
That is fast to review

carlosclaiton (commented):
Should you comment out the whole check, instead?

FaceWithDark:
> Should you comment out the whole check, instead?

The check was there to deny processing query received from user if no active services found so...

carlosclaiton (approved):
pass on the test

PR #143, opened 2026-06-10
Cosmic 285: Metadata Generation

marslanshaukat:
This PR is for

1. Disabling the auto indexation of files at the start 
2. Metadata generation and saving to document_index.json 
3. Multiple uploads correctly append (do not overwrite) existing entries

Metadata includes required fields:
    document_id
    user_id
    session_id
    file_name
    memory_type
    chunk_count
    upload_date


Adnan525:
I can see the 2 new classes - DocumentMetadata and DocumentIndex but they were never mentioned/used anywhere else. I added breakpoints in those classes to double-check what metadata it is generating and there's nothing. can you double check this and am i missing something?

src/services/document_index.py
Adnan525, line 1: I can see it generates metadata for each uploaded file but the filtering is never used. dont we need to filter during retrieval?

src/opensi_cosmic.py
Adnan525, line 1: Image

global docs like academic governance is present but not indexed in document_index.json. is this on purpose?

marslanshaukat: Yes. That functionality has been disabled to avoid loading at startup. In future, any information for global memory would be handled by memory variable global_memory.

src/services/document_index.py
marslanshaukat: Yes. This PR is only for metadata generation. Filtering will be implemented in future PR. Thanks.

src/opensi_cosmic.py
carlosclaiton: Should the global memory to initiate on the start up? Like it should move it for the correct location now in this PR and laod and index the file at the vb?

backend/routers/memory.py
carlosclaiton: even if the memory is global, we do require it do be done by services, as each service would have a different set of files and you will require to know the information to filter later on.

carlosclaiton (changes requested):

src/opensi_cosmic.py
marslanshaukat: For a fresh start, no indexation will take place to reduce system loading time. The idea is that once the system is up, the admin will add documents to global_memory. This information will persist even if the entire system is restarted. Thanks.

backend/routers/memory.py
marslanshaukat: During our last meeting, we agreed to keep only global memory for now because adding too much information may cause the LLM to hallucinate in its responses. Current implementation of metadata can attach memory_type for each file but then we need to think how retrieval architecture will work, such as which service should take priority during retrieval when there are multiple services involved and so on.

carlosclaiton: In our last meeting we agreed that we need to have folder for each service that has memory. For example, academic_governance. So we need a folder to store the documents that now are in the docs folder.

src/opensi_cosmic.py
```diff
@@ -355,16 +355,16 @@ def set_up_qa(
...
+            #     follow_symlinks=True
+            # ):
+            #     vector_database.add_document_directory(self.config_data["doc_directory"])
 
-            # Add documents.
-            if self.config_data["document_path"] != "" or len(self.config_data["document_path"]) > 0:
-                vector_database.add_documents(self.config_data["document_path"])
+            # # Add documents.
+            # if self.config_data["document_path"] != "" or len(self.config_data["document_path"]) > 0:
+            #     vector_database.add_documents(self.config_data["document_path"])
```
carlosclaiton, line 367: this line should still happen at the beginning with the following logic, if a service that requires internal memory is active, it should index the content from the folder /app/data/memories/global/{service_name}

marslanshaukat: I will enable vector embedding indexation in future PR

carlosclaiton (approved):

PR #145, opened 2026-06-12
Use **Configurations API** endpoints from `COSMIC-DB` instead of existing one [COSMIC-225]

FaceWithDark:
*Ref: [COSMIC-225 ticket](https://opensi.atlassian.net/browse/COSMIC-225?atlOrigin=eyJpIjoiOTM0ZmY3ZTMwYWVjNGI3MzgzZmRhMGI4NmU0OWZjY2MiLCJwIjoiaiJ9)*

> [!NOTE]
> Make sure [#44](https://github.com/TheOpenSI/CoSMIC_DB/pull/44) (`CoSMIC_DB` repo) merged first. Otherwise, someone's machine will exploded **literally** since the current default SLM for QA is `llama3.3:70b`😃
- I did change from `is_quantized` &rarr; `is_quantised` as **Australian English** mostly adopt from **British English**. *Probably this's not needed*


src/query_analyser/query_analyser.py
```diff
@@ -25,19 +25,28 @@ def __init__(
         self,
         llm_name:       str     = "mistral-7b-instruct-v0.1",
```
carlosclaiton: Can we change to the default to be qwen2.5:7b instead

src/opensi_cosmic.py
```diff
@@ -43,43 +43,67 @@
 class OpenSICoSMIC:
     def __init__(
         self,
-        query_llm_name: str         = "",
-        llm_name:       str         = "",
-        config_path:    str         = "scripts/configs/config.yaml",
+        general_slm:    str         = "",
+        qa_slm:         str         = "",
+        config_path:    str         = "",
```
carlosclaiton: is that still required here?

src/opensi_cosmic.py
```diff
@@ -43,43 +43,67 @@
 class OpenSICoSMIC:
     def __init__(
         self,
-        query_llm_name: str         = "",
-        llm_name:       str         = "",
-        config_path:    str         = "scripts/configs/config.yaml",
+        general_slm:    str         = "",
```
carlosclaiton, line 45: This one is the SLM used for the query analyser, not general questions.

src/opensi_cosmic.py
```diff
@@ -43,43 +43,67 @@
...
+        qa_slm:         str         = "",
```
carlosclaiton, line 46: this is used for general questions, qa here does not stand for query analyser.

src/opensi_cosmic.py
```diff
@@ -276,13 +327,13 @@ def __call__(
             ) = self.qa(
                 query=question,
                 services=self.get_services(),
-                # The context that we are passing here is chat history. See api.py
-                context=context,
+                context=context, # Chat history context (see `backend/routers/cosmic.py`)
                 is_rag=True,
                 verbose=False
```
carlosclaiton, line 363: This should be reading from the environment variable in a way we can change it without have to build and get logs

src/opensi_cosmic.py
```diff
@@ -348,31 +399,36 @@ def set_up_qa(
...
+            # another branch right now. However, I need to match what already
+            # there in the YAML file so COSMIC-225 PR can be merged. Once this
+            # merged, we can modify the path again with the current branch
+            # working on RAG to test out.
+
             vector_database = VectorDatabase(
-                document_analyser_model="gte-small",
-                local_database_path=str(object=vector_db_path),
-                vector_database_update_threshold=0.98,
+                local_database_path="",
```
carlosclaiton, line 441: where is the information of the embedding model?

src/opensi_cosmic.py
carlosclaiton: After reading all the code, I noticed that you seem to have made the change in a consistent way. I will test this as soon as I manage to get a computer with the minimum requirements.

src/opensi_cosmic.py
carlosclaiton: same as above.

carlosclaiton (commented):
I still need to test, but here are a few comments that you could already start addressing.

src/opensi_cosmic.py
FaceWithDark: I swapped params order around:

- `qa_slm` was `query_llm_name`
- `general_slm` was `llm_name`

The reason I named it as `general_slm` because it's coming from the **General config**, not because I'm referring to the **General Questions Answering** service.

src/opensi_cosmic.py
FaceWithDark: Similar reasoning as [R46](https://github.com/TheOpenSI/CoSMIC/pull/145/changes#r3417590982).

src/opensi_cosmic.py
FaceWithDark: We can remove it. I wasn't sure if we wanted to support alternative way to providing custom configs by a particular file format (e.g., `.yaml`) or not.

src/opensi_cosmic.py
FaceWithDark: I don't quite understand your intention by this review. Could I've more details on it?

src/opensi_cosmic.py
FaceWithDark: Seems like @marslanshaukat did provide default value to some params in `VectorDatabase()` class (source from `src/services/vector_database.py`) so I just omitted it with the exception for `local_database_path` to match with what we had previously in `scripts/configs/config.yaml`.

src/query_analyser/query_analyser.py
FaceWithDark: Yes, the change can be view from commit 4b5cbe67.

FaceWithDark:
> I still need to test, but here are a few comments that you could already start addressing.

Thank you. I've mark resolved to one of the review. For the rest, I can't resolved just yet due to various reasons. Please refer to my statement on those aren't yet marked.

carlosclaiton:
Also, I think we need to modify the readme, as it now does not require the openai_api_key in the .env, should be safe to keep this key at the database?

src/opensi_cosmic.py
carlosclaiton: That verbose is set as False by default, I am suggesting verbose should read from .env, that way we can set it to true in a single place and use that to build the dev mode, where it prints lots of logs to help debugging.

carlosclaiton: we can do that on another PR

FaceWithDark: Ohh that one. Yeah, I think another PR is better since we should apply the same thing to other classes/functions that have `verbose` (or should have) param as well.

FaceWithDark:
> Also, I think we need to modify the readme, as it now does not require the openai_api_key in the .env, should be safe to keep this key at the database?

It doesn't have any effects to CoSMIC anymore if we does/still have the variable defined in the `.env` at the moment. I'll just update to remove it from `README`.

carlosclaiton:
One more thing. If I modify the config through the front end, it is not making effect on cosmic. I set the query analyser with a different model, it used the same models for both.

carlosclaiton:
> One more thing. If I modify the config through the front end, it is not making effect on cosmic. I set the query analyser with a different model, it used the same models for both.

This one, probably the front end should rebuild opensi-cosmic when updating the database, would you think that is the best way?

FaceWithDark:
> One more thing. If I modify the config through the front end, it is not making effect on cosmic. I set the query analyser with a different model, it used the same models for both.

Hmm... interesting. I removed the config checking logic in `/cosmic` endpoint because I thought it'd work perfectly out of the box. I do know that when we update the config on **config page**, it sent `PATCH` request to our API endpoint. This means CoSMIC can detect the changes near instant since for every request to `/cosmic` endpoint, it calls the `OpenSICoSMIC()` class that have method to fetch latest config data everytime.

I'm testing it now. Maybe, I'll find out why it doesn't work like how you and me expected.

carlosclaiton:
Screenshot 2026-06-16 at 1 25 29 pm

Note that the INFO is not displaying the correct model for QA SLM

FaceWithDark:
Okay, I know why now. It's stale reference issue, quite similar to how I found the issue in #132. The previous check we've in `backend/routers/cosmic.py` should be kept but the logic needs to be modified a bit since we don't check based on the timestamp of the YAML file anymore.

carlosclaiton:
> Okay, I know why now. It's stale reference issue, quite similar to how I found the issue in #132. The previous check we have in `backend/routers/cosmic.py` should be kept, but the logic needs to be modified a bit since we don't check based on the timestamp of the YAML file anymore.

I also noticed that we have the ability to update the config using the id or it just creates a new one. We should have a look at this, as it can cause issues. 

We can always look for the new time stamp and keep track to rebuild if the config has changed, or we could call a rebuild when saving the new config from the front end. (This one we should have a API end point to rebuild cosmic)

carlosclaiton:
We also needs to check, as I did rebuild cosmic using two different models, but it seems it is creating all the responses using only the query analyser model.

carlosclaiton:
> We also needs to check, as I did rebuild cosmic using two different models, but it seems it is creating all the responses using only the query analyser model.

This is based on the quality of the answers, as I could not figure out how to know what model is used through the logs from the ollama container.

FaceWithDark:
> > Okay, I know why now. It's stale reference issue, quite similar to how I found the issue in #132. The previous check we have in `backend/routers/cosmic.py` should be kept, but the logic needs to be modified a bit since we don't check based on the timestamp of the YAML file anymore.
> 
> I also noticed that we have the ability to update the config using the id or it just creates a new one. We should have a look at this, as it can cause issues.
> 
> We can always look for the new time stamp and keep track to rebuild if the config has changed, or we could call a rebuild when saving the new config from the front end. (This one we should have a API end point to rebuild cosmic)

I'm thinking of 2 ways to solve this:

1. `State()` the default config and check for update to rebuild (if needed) using the old check we have in `backend/routers/cosmic.py`. This works similar to how we de-double initialising `OpenSICoSMIC()` calls in #132.

2. Have another `/cosmic` endpoint but accepts `PATCH` request. This way, when FE updates the config, it'd call both the **API endpoint** _(to update in the db)_ & **this new endpoint** _(to trigger `OpenSICoSMIC()` for a rebuild)_.

Which one would you prefer? I'm thinking **Option 2** is better to cleanly separate the business logic (`/cosmic` endpoint [`POST` request]) and other responsibilities. This plays well to [RESTful API](https://www.ibm.com/think/topics/rest-apis) design too.

carlosclaiton:
> > > Okay, I know why now. It's stale reference issue, quite similar to how I found the issue in #132. The previous check we have in `backend/routers/cosmic.py` should be kept, but the logic needs to be modified a bit since we don't check based on the timestamp of the YAML file anymore.
> > 
> > 
> > I also noticed that we have the ability to update the config using the id or it just creates a new one. We should have a look at this, as it can cause issues.
> > We can always look for the new time stamp and keep track to rebuild if the config has changed, or we could call a rebuild when saving the new config from the front end. (This one we should have a API end point to rebuild cosmic)
> 
> I'm thinking of 2 ways to solve this:
> 
> 1. `State()` the default config and check for update to rebuild (if needed) using the old check we have in `backend/routers/cosmic.py`. This works similar to how we de-double initialising `OpenSICoSMIC()` calls in [Fix double initiallisation issue during development on `CoSMIC` repo [COSMIC-229] (New) #132](https://github.com/TheOpenSI/CoSMIC/pull/132).
> 2. Have another `/cosmic` endpoint but accepts `PATCH` request. This way, when FE updates the config, it'd call both the **API endpoint** _(to update in the db)_ & **this new endpoint** _(to trigger `OpenSICoSMIC()` for a rebuild)_.
> 
> Which one would you prefer? I'm thinking **Option 2** is better to cleanly separate the business logic (`/cosmic` endpoint [`POST` request]) and other responsibilities. This plays well to [RESTful API](https://www.ibm.com/think/topics/rest-apis) design too.

I prefer the new post request api endpoint. I think that is more elegant and will be more efficient than checking for each question if the configuration has changed. 

[COSMIC-229]: https://opensi.atlassian.net/browse/COSMIC-229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

FaceWithDark:
> It doesn't have any effects to CoSMIC anymore if we does/still have the variable defined in the `.env` at the moment. I'll just update to remove it from `README`.

I've updated this as of commit 01b0ce8.


> I prefer the new post request api endpoint. I think that is more elegant and will be more efficient than checking for each question if the configuration has changed.

I've added this as of commit ad78aed. To test, you can do the following steps:

1. Go to **config** page and modify any default value for either configs, then hit `Save`.

2. We didn't have the new code on FE now to automatically trigger the new `/cosmic` endpoint. Therfore, please go to [localhost:3000/docs](http://localhost:3000/docs) and manually trigger it. You should see the different between the 1st & 2nd logs.

3. _(Optional)_ Try asking CoSMIC some quesitons before and after modifying configs. You don't have to reload the current page or delete the current chat. It's absolutely just fine to perform this on the same brower session **(this's how I tested as well)**. `I also added more logging if you test this optional step as well`.


> I also noticed that we have the ability to update the config using the id or it just creates a new one. We should have a look at this, as it can cause issues.

Erm... I don't quite get this one yet. Yes, we can create config presets (refer to my comments in the codebase) but when updating, we must know the **configs ID (UUID in this case)** in order to execute a valid `PATCH` request. I did setup some strict validations in the endpoint so we can ensure that the right **configs `JSONB` data** get sent to the selected **configs ID** in the db.

Could I've more info on this matter?


> > We also needs to check, as I did rebuild cosmic using two different models, but it seems it is creating all the responses using only the query analyser model.
> 
> This is based on the quality of the answers, as I could not figure out how to know what model is used through the logs from the ollama container.

I'll give a diagnose regarding this interesting *bug* (I'd say). To my understanding, QA should use the `General SLM` on 1st SLM call, while on 2nd SLM call is the `QA SLM` right?

carlosclaiton:
> > It doesn't have any effects to CoSMIC anymore if we does/still have the variable defined in the `.env` at the moment. I'll just update to remove it from `README`.
> 
> I've updated this as of commit [01b0ce8](https://github.com/TheOpenSI/CoSMIC/commit/01b0ce8499397de00b39e862f14368bc4557d107).
> 
> > I prefer the new post request api endpoint. I think that is more elegant and will be more efficient than checking for each question if the configuration has changed.
> 
> I've added this as of commit [ad78aed](https://github.com/TheOpenSI/CoSMIC/commit/ad78aed785a19cdd1c43255c9dcc1527e1bd51e3). To test, you can do the following steps:
> 
> 1. Go to **config** page and modify any default value for either configs, then hit `Save`.
> 2. We didn't have the new code on FE now to automatically trigger the new `/cosmic` endpoint. Therfore, please go to [localhost:3000/docs](http://localhost:3000/docs) and manually trigger it. You should see the different between the 1st & 2nd logs.
> 3. _(Optional)_ Try asking CoSMIC some quesitons before and after modifying configs. You don't have to reload the current page or delete the current chat. It's absolutely just fine to perform this on the same brower session **(this's how I tested as well)**. `I also added more logging if you test this optional step as well`.
> 
> > I also noticed that we have the ability to update the config using the id or it just creates a new one. We should have a look at this, as it can cause issues.
> 
> Erm... I don't quite get this one yet. Yes, we can create config presets (refer to my comments in the codebase) but when updating, we must know the **configs ID (UUID in this case)** in order to execute a valid `PATCH` request. I did setup some strict validations in the endpoint so we can ensure that the right **configs `JSONB` data** get sent to the selected **configs ID** in the db.
> 
> Could I've more info on this matter?
> 
> > > We also needs to check, as I did rebuild cosmic using two different models, but it seems it is creating all the responses using only the query analyser model.
> > 
> > 
> > This is based on the quality of the answers, as I could not figure out how to know what model is used through the logs from the ollama container.
> 
> I'll give a diagnose regarding this interesting _bug_ (I'd say). To my understanding, QA should use the `General SLM` on 1st SLM call, while on 2nd SLM call is the `QA SLM` right?

Actually the first question is answered by the query analyser, it first detected what is the service and if the question is about the system, only when it is other services it should use the general questions.

src/opensi_cosmic.py
FaceWithDark: I think we can just remove it for now. If someone requested for this feature and it's actually a needed thing for many people, then we can added it back. Changes are trackable at commit c9eee2c.

FaceWithDark:
> Actually the first question is answered by the query analyser, it first detected what is the service and if the question is about the system, only when it is other services it should use the general questions.

Okay, so I've tested and saw QA called SLMs 3 times to answer my query - `Yo CoSMIC! How can you help me?`. I guess that normally, it should do only 2 times but because my query triggered the system prompt (see the log below), it asked itself again (with `QA SLM`). Therefore, I believe it does perform like how you said in the quoted comment in this comment.

For my machine and my test case, I used `llama3.2:1b` (**General config**) & `mistral:latest` (**QA config**) because I can only run on CPU with my Windows machine. However, you can use any SLMs to test this.


Relevant test logs:
```md
### 'cosmic-fastapi' container ###


## This's after my query get:
## 1. Sent to CoSMIC (`OpenSICoSMIC.__call__()`)
## 2. Sent to QA (`QueryAnalyser.__call__()`)
## 3. Return to FE (`result: <response>`) for display


[Info] Default General configs (during execution): {'provider': 'ollama', 'model': 'llama3.2:1b', 'is_quantised': False, 'seed': 0, 'default_knowledge_path
': '/app/data/default/', 'temp_knowledge_path': '/app/data/temp/', 'api_key': ''}
[Info] Default QA configs (during execution): {'provider': 'ollama', 'model': 'mistral:latest', 'is_quantised': False, 'seed': 0, 'default_knowledge_path':
 '/app/data/default/', 'temp_knowledge_path': '/app/data/temp/', 'api_key': ''}
[Info] Set General SLM (during execution): [ollama:llama3.2:1b]
[Info] Set QA SLM (during execution): [ollama:mistral:latest]
Ollama models used: mistral:latest
Ollama models used: mistral:latest
[Info] Does SLM response detected user query asking about our system info or not? ( YES)
[Info] System info trigged: (True)
Ollama models used: llama3.2:1b
[Info] [Emissions DB] Successfully stored emissions for user 019eceab-c5c4-71cc-8343-3ced3bd69c4e
[Info] [CodeCarbon] General chat query emissions: 0.00046188 kg CO₂
INFO:     172.18.0.1:53524 - "POST /api/v1/cosmic HTTP/1.1" 200 OK
```


```md
### 'cosmic-ollama' containter ###


## Block of logs here... ##


# Saw Ollama starting loading in 'QA SLM' first
llama_model_loader: - kv   0:                       general.architecture str              = llama
llama_model_loader: - kv   1:                               general.name str              = Mistral-7B-Instruct-v0.3


# Optional lookup
print_info: general.name          = Mistral-7B-Instruct-v0.3


## Continue block of logs here... ##


# Not too sure but these line seems like the way Ollama "remember" previous context if called the 
# same model again. For our case, this means `QA SLM` performed 2nd call based on context of 
# 1st call.
slot get_availabl: id  0 | task -1 | selected slot by LCP similarity, sim_best = 0.594 (> 0.100 thold), f_keep = 0.397
srv  get_availabl: updating prompt cache
srv   prompt_save:  - saving prompt with length 262, total state size = 32.754 MiB (draft: 0.000 MiB)
srv          load:  - looking for better prompt, base f_keep = 0.397, sim = 0.594
srv        update:  - cache state: 1 prompts, 32.754 MiB (limits: 8192.000 MiB, 4096 tokens, 65528 est)
srv        update:    - prompt 0x8f7f650:     262 tokens, checkpoints:  0,    32.754 MiB
srv  get_availabl: prompt cache update took 92.80 ms


## Continue block of logs here... ##


# Saw Ollama starting loading in 'General SLM' second
llama_model_loader: - kv   0:                       general.architecture str              = llama
llama_model_loader: - kv   1:                               general.name str              = Llama 3.2 1B Instruct


# Optional lookup
print_info: general.name          = Llama 3.2 1B Instruct


## Continue block of logs here... ##


# Looks like the same pattern used here. Ollama only "see" `General SLM` called once so no
# previous context needed to "remember".
slot get_availabl: id  0 | task -1 | selected slot by LRU, t_last = -1
srv  get_availabl: updating prompt cache
srv          load:  - looking for better prompt, base f_keep = -1.000, sim = 0.000
srv        update:  - cache state: 0 prompts, 0.000 MiB (limits: 8192.000 MiB, 4096 tokens, 8589934592 est)
srv  get_availabl: prompt cache update took 0.01 ms


## Continue block of logs here... ##
```


I've added more debug msgs (commit be0c08e) so you can track down and see the same output like mine.

FaceWithDark:
> [!NOTE]
> Also, just like #142, I'll remove those extra debug msgs when the PR is approved.

carlosclaiton (approved):
I have tested the new rebuild endpoint by modifying the models in the front end and calling the api served on 3000. It does detected when it has new configuration and rebuild it. 

I also liked the new debbug message to test calling order. 

```md
[Info] Default General configs (during execution): {'provider': 'ollama', 'model': 'mistral:7b', 'is_quantised': False, 'seed': 0, 'default_knowledge_path': '/app/data/default/', 'temp_knowledge_path': '/app/data/temp/', 'api_key': ''}
[Info] Default QA configs (during execution): {'provider': 'ollama', 'model': 'qwen2.5:7b', 'is_quantised': False, 'seed': 0, 'default_knowledge_path': '/app/data/default/', 'temp_knowledge_path': '/app/data/temp/', 'api_key': ''}
[Info] Set General SLM (during execution): [ollama:mistral:7b]
[Info] Set QA SLM (during execution): [ollama:qwen2.5:7b]
[Info] Ollama models used: qwen2.5:7b
[Info] Ollama models used: qwen2.5:7b
[Info] Does SLM response detected user query asking about our system info or not? (YES )
[Info] System info trigged: (True)
[Info] Ollama models used: mistral:7b
[Info] [Emissions DB] Successfully stored emissions for user 019ed9bf-79a9-7321-bda4-50318840dcfe
[Info] [CodeCarbon] General chat query emissions: 0.00024669 kg CO₂
INFO:     172.18.0.1:37462 - "POST /api/v1/cosmic HTTP/1.1" 200 OK
```

I have solved the conflict with the new CodeCarbon implementation to ensure it does update the information on the database

Todo: ensure front end calls the new API endpoint when saving new configuration.

FaceWithDark:
> Note
> 
> Also, just like #142, I'll remove those extra debug msgs when the PR is approved.

Disabled as of commit 8737d9f.

> Todo: ensure front end calls the new API endpoint when saving new configuration.

I'll create a ticket for this shortly.

carlosclaiton:
It seems you made more changes than disable the prints. Sent from my iPhoneOn 18 Jun 2026, at 7:29 pm, DeepInDark ***@***.***> wrote:﻿FaceWithDark left a comment (TheOpenSI/CoSMIC#145)

Note
Also, just like #142, I'll remove those extra debug msgs when the PR is approved.

Disabled.

FaceWithDark:
> It seems you made more changes than disable the prints. Sent from my iPhoneOn 18 Jun 2026, at 7:29 pm, DeepInDark ***@***.***> wrote:﻿FaceWithDark left a comment ([TheOpenSI/CoSMIC#145](https://github.com/TheOpenSI/CoSMIC/pull/145)) Note Also, just like #142, I'll remove those extra debug msgs when the PR is approved. Disabled. —Reply to this email directly, view it on GitHub, or unsubscribe.Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS and Android. Download it today! You are receiving this because your review was requested.Message ID: ***@***.***>

I just added type-hints and re-format the docs for methods inside classes that got modified in this PR. Mainly just linter stuffs 😅.

src/query_analyser/query_analyser.py
```diff
@@ -98,7 +98,7 @@ def __call__(
         # update to handle `int` properly
         services:   dict[str, dict[str, str]],
         verbose:    bool = False
-    ):
+    ) -> tuple[str, dict[str, str | bool]]:
```
carlosclaiton, line 101: this is not correct information, I think verbose is not part of the dictionary.

FaceWithDark: It's the boolean from `service_info_dict` dictionary. I saw we return `service_option` & `service_info_dict` in `__call__()` func, which the dictionary contains:

- User query that trigger the system prompt
- The trigger indicator
- The system prompt itself

You can check on where the dictionary defined on **line 185**.

PR #148, opened 2026-06-17
Cosmic 195 codecarbon

sukeerthreddy12:
made sure the Api connection is established and removed the csv integration as we no longer needed it and successfully stores data in emissions table in DB


carlosclaiton (approved):
it is working

PR #152, opened 2026-06-25
 feat(RAG): Unify metadata payloads and scoped retrieval, watchdog po…

marslanshaukat:
This PR unifies document metadata across the JSON index and Qdrant payloads, embeds ingested files into the vector store, and adds scoped retrieval so session, user, and global documents can safely share one collection.

1. Added enriched document metadata fields, including content hash, file size, file extension, content type, and updated date.
2. Added a single payload builder so JSON index rows and vector payloads stay in sync.
3. Embedded documents on ingestion for both global folder drops and API uploads.
4. Attached full metadata to every Qdrant chunk and fixed vector titles to use the original file name.
5. Added scoped retrieval filters for session, user, and global memory tiers.
6. Added hash-based change detection and overwrite/reindex behavior for modified files.
7. Added stale vector cleanup on delete and reindex paths.
8. Improved watcher behavior so runtime file drops are picked up reliably.
9. Fixed file-chat handling so attached-file messages return a valid answer body.

Note:
Saving and retrieval for user memory is not fully functional. We need to think how to address this issue probably a new call to LLM to check with some classifiers. 

**COSMIC-286 must be tested with FE: COSMIC-287**


carlosclaiton (commented):
Why it is showing 28 files to change, it should not require that much change. 

I can see why saving into user memory will require a special architecture, which is our old implementation of service called memory. Although I can not see why it would have a retrieval problem, as we discussed it should always look at it to retrieve information related to the user.

backend/routers/cosmic.py
```diff
@@ -67,6 +69,13 @@ class CosmicAPI(BaseModel):
     user_message:   str
     body:           Body
 
+
+# Initialize document indexes for metadata tracking
+global_document_index = DocumentIndex(persist_path="/app/data/memories/global/global_document_index.json")
+user_document_index = DocumentIndex(persist_path="/app/data/memories/users/user_document_index.json")
+memory.set_document_indexes(global_document_index, user_document_index)
+
```
carlosclaiton: would that happens everytime this api will be called? should it check if documents has been changed?

backend/routers/cosmic.py
```diff
@@ -217,7 +226,14 @@ async def process_cosmic(
...
                 # Update vector database
-                answer: str = str(opensi_cosmic(question=user_message_vector_db_update)[0])
+                answer: str = str(
+                    opensi_cosmic(
+                        question=user_message_vector_db_update,
+                        session_id=data.chat_id,
+                        has_files=True,
+                        user_id=user_id,
+                    )[0]
+                )
```
carlosclaiton: this is active only when the question request for memory of the user? It is not clear to me what is the logic here.

.env.example
```diff
@@ -0,0 +1,47 @@
+
+# cp .env.example .env
+# .env is gitignored; only this template is committed. Values shown are the
+# in-code defaults, so an unset variable falls back to the same behaviour.
+
+# Docker service name / host for the Ollama server.
+OLLAMA_SERVICE_NAME=ollama
```
carlosclaiton: cosmic-ollama

That will match with all the others containers

.env.example
carlosclaiton, line 1: most of this should move to database RAG configuration

.env.example
```diff
@@ -0,0 +1,47 @@
...
+
+# Vector database
+QDRANT_URL=http://qdrant:6333
+
+# Embedding model
+RAG_EMBEDDING_MODEL=gte-small
```
carlosclaiton, line 13: thought we agreed that we would use the qwen enbedding model or another one that was better than this one.

backend/routers/cosmic.py
carlosclaiton: as you created the .env_example you could add this one to it as well

carlosclaiton: Actually create a JIRA ticket to it and a PR for this modification.

src/services/document_index.py
```diff
@@ -0,0 +1,377 @@
...
+            "document_id": self.document_id,
+            "file_name": self.file_name,
+            "memory_type": self.memory_type,
+            "chunk_count": self.chunk_count,
+            "upload_date": self.upload_date,
+            "updated_date": self.updated_date,
+            "content_hash": self.content_hash,
+            "file_size": self.file_size,
+            "file_ext": self.file_ext,
+            "content_type": self.content_type,
```
carlosclaiton, line 78: what is content_type?

src/services/document_index.py
```diff
@@ -0,0 +1,377 @@
...
+        self.file_size = file_size
+        self.file_ext = file_ext or (os.path.splitext(file_name)[1].lower() if file_name else "")
+        self.content_type = content_type
+        self.updated_date = updated_date or self.upload_date
+
+    @property
+    def title(self) -> str:
+        return os.path.splitext(self.file_name)[0] if self.file_name else ""
+
+    def to_dict(self) -> Dict[str, Any]:
```
carlosclaiton, line 67: we need a PR to move the storage of the metadata into the database, as this should be kept in a secure location in case needs to rebuild cosmic container.

src/services/qa.py
```diff
@@ -231,25 +238,22 @@ def __call__(
             raw_response    = f"The next moves are from {next_move}.\n{raw_response}"
 
         elif service_option == "1":
```
carlosclaiton: service 1 is the memory update,this one is related to user memory not service name. Then why do you need to get the service name?

src/services/qa.py
```diff
@@ -258,10 +262,10 @@ def __call__(
             )= self.code_generator(query)
 
         else:
-            RAG_ENABLED_SERVICES = ["4"] # Academic QA triggers retrieval
+            RAG_ENABLED_SERVICES = ["3", "4"]
```
carlosclaiton: that needs to read from the database, as we can not edit the core code everytime we have a new service. And RAG enable is bad name as it should be to know if the service has memory associate as all service has RAG as it should be able to get information from documents as context if a user upload a document.

src/services/qa.py
```diff
@@ -258,10 +262,10 @@ def __call__(
...
             execute_rag = (
                 (is_rag) and
-                (service_option in RAG_ENABLED_SERVICES)
+                (service_option in RAG_ENABLED_SERVICES or has_files)
```
carlosclaiton, line 311: here the item has_files make that service 3 should not be required on the list above

src/services/vector_database.py
```diff
@@ -26,84 +25,37 @@ def __init__(
         self,
         document_analyser_model: str = "gte-small",
         # document_analyser_model: str = "Qwen/Qwen3-Embedding-8B",
```
carlosclaiton, line 27: here is the model I was interested in testing

src/opensi_cosmic.py
```diff
@@ -854,3 +848,128 @@ def get_configs(
...
+
+# Service Registry Utilities (for StorageEventWatcher and memory management)
+
+_service_cache = {} # To avoid hitting the backend API frequently
+_service_cache_timestamp = 0
+_CACHE_TTL = 30  # seconds; short TTL so service status/capability toggles take effect promptly
+
+
+def _get_raw_services(
+    url: str = "http://backend:8000/api/v1/services/",
```
carlosclaiton: this part should read from the .env combine with /v1/services

carlosclaiton (commented):
looking good progress, I still have a few questions that need to be answered before be able to merge. 

Also we do need to stress test this, have you add a few documents into the memory of service, user and chatsession and check if the retrieved information is correct?

backend/routers/cosmic.py
marslanshaukat: Moved to api to avoid double initialization. 

This will also check if document is modified.

src/opensi_cosmic.py
```diff
@@ -854,3 +848,128 @@ def get_configs(
...
+            response.raise_for_status()
+            services = response.json().get("result", [])
+            _service_cache = services
+            _service_cache_timestamp = current_time
+            return services
+    except Exception as e:
+        # Return cached data if available, otherwise empty list
+        if _service_cache:
+            return _service_cache
+        return [5]
```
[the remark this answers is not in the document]
marslanshaukat: hard-coded to fall back to academic governance

src/services/qa.py
```diff
@@ -231,25 +238,22 @@ def __call__(
...
-
-                if text is not None:
-                    # Add text to database.
-                    self.rag.vector_database.update_database_from_text(text=text)
-
-            response = raw_response = "Vector database updated."
+            selected_service_name = services_name.get(service_option, "").lower()
+            if (
+                memory_service_active
+                and selected_service_name == "memory"
```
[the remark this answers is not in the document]
marslanshaukat: Created a new ticket to enhance memory logic

backend/routers/cosmic.py
```diff
@@ -67,6 +69,13 @@ class CosmicAPI(BaseModel):
     user_message:   str
     body:           Body
 
+
+# Initialize document indexes for metadata tracking
+global_document_index = DocumentIndex(persist_path="/app/data/memories/global/global_document_index.json")
+user_document_index = DocumentIndex(persist_path="/app/data/memories/users/user_document_index.json")
+memory.set_document_indexes(global_document_index, user_document_index)
```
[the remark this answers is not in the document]
marslanshaukat: moved to api to avoid double initialization

backend/routers/cosmic.py
marslanshaukat: You are right we don't need this logic here anymore

.env.example
marslanshaukat: I think for development its best to keep

.env.example
marslanshaukat: Created a new ticket for RAG retrieval enhancement and this will be addressed there

src/services/document_index.py
marslanshaukat: its store which application the extension belongs to. For example .pdf ext renamed to .jpg this helps to future feature of file validation. This will especially be helpful when using agent for tool calling so that it knows that which application that ext belongs to.

src/services/document_index.py
marslanshaukat: Created a new ticket

marslanshaukat:
Most of the issues are addressed and resolved except related to user memory (qa.py). I have noted down the comments and created a new ticket for user memory enhancement.

carlosclaiton:
@marslanshaukat this PR has some issues with permission, I always have to pull it as sudo otherwise:

Updating c981c65..1a346d7
Fast-forward
fatal: update_ref failed for ref 'HEAD': cannot update the ref 'refs/heads/COSMIC-286': unable to append to '.git/logs/refs/heads/COSMIC-286': Permission denied

Can you please make sure it is set as the others and put back all the files for the chess example.

src/opensi_cosmic.py
```diff
@@ -854,3 +848,129 @@ def get_configs(
...
             )
+
+
+# Service Registry Utilities (for StorageEventWatcher and memory management)
+
+_service_cache = {} # To avoid hitting the backend API frequently
+_service_cache_timestamp = 0
+_CACHE_TTL = 30  # seconds; short TTL so service status/capability toggles take effect promptly
+
+SERVICES_API_URL = os.getenv("SERVICES_API_URL","http://backend:8000/api/v1/services/")
```
carlosclaiton: You should avoid defining an environment variable for each API endpoint. Instead, define a single base URL for the backend server and construct endpoint-specific URLs in code.

API_BASE_URL = os.getenv("API_BASE_URL", "http://localhost:8000")
SERVICES_API_URL = f"{API_BASE_URL}/api/v1/services/"

src/opensi_cosmic.py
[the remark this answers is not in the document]
carlosclaiton: second copilot it will break, and I agree if it.

.env.example
```diff
@@ -0,0 +1,53 @@
...
+RAG_RERANK_SCORE_THRESHOLD=0.0
+
+# Base URL the pipeline uses to reach the CoSMIC API.
+OPENSI_COSMIC_API_BASE_URL=http://cosmic:3000
+
+# Max queries per non-admin user.
+MAX_QUERIES_PER_USER=5
+
+# Chat API
+CHAT_API_URL=http://backend:8000/api/v1/chatboxes/
```
carlosclaiton, line 50: as mentioned on my comment above, this should be only one env variable API_DATABASE_URL, and inside of the code you build the API end point

marslanshaukat: Should we create a config helper file to build a URL path instead? This will help us to make changes in a single file. For example in .env we will set base url and config will build constant and in future lets say it changes to  v2/chatboxes, we will only need to change it at one place.

marslanshaukat:
Now can chat with uploaded file and file chip is attached in the FE
image

src/opensi_cosmic.py
```diff
@@ -854,3 +850,129 @@ def get_configs(
...
+
+def _get_raw_services(
```
carlosclaiton: This function is duplicating the fucntionality of function get_services

src/opensi_cosmic.py
```diff
@@ -854,3 +850,129 @@ def get_configs(
...
+            _service_cache_timestamp = current_time
+            return services
+    except Exception as e:
+        # Return cached data if available, otherwise empty list
+        if _service_cache:
+            return _service_cache
+        return [5]
+
+
+def get_service_id_by_name(service_name: str) -> int | None:
```
carlosclaiton: you can read this from the dictionary returned by get_services function

src/opensi_cosmic.py
```diff
@@ -854,3 +850,129 @@ def get_configs(
...
+    # If no services marked with memory_capability, use hardcoded fallback
+    if not rag_services:
+        return [5]  # academic_governance
+
+    return rag_services
+
+def get_active_memory_scope() -> tuple[list[str], bool]:
+    """ Resolve global memory scope and whether the user-memory service is active """
+    
+    services = _get_raw_services() # service is only used when active
```
carlosclaiton: use get_service instead

src/services/qa.py
carlosclaiton: It is not longer the correct index

backend/routers/cosmic.py
carlosclaiton, line 1: Why is this file showing so much change, but the functions are just moving around?

PR #156, opened 2026-07-02
fix: hard-coded `user_role` label value instead of using user's role

FaceWithDark:
> [!IMPORTANT]
> Refer to my [backtracking process](https://github.com/TheOpenSI/CoSMIC_DB/pull/52#issuecomment-4862091944) on [#52](https://github.com/TheOpenSI/CoSMIC_DB/pull/52) (`CoSMIC_DB` repo) for more details. This one must be merged first so that mentioned PR can also be merge with functional changes.

### DIFF
--- a/backend/routers/cosmic.py
+++ b/backend/routers/cosmic.py
@@ -140,7 +140,7 @@ async def process_cosmic(
 ):
     try:
         user_id:    str     = data.body.model_dump(mode="json")["user"]["id"]
-        user_role:  str     = data.body.model_dump(mode="json")["user"]["role"]
+        # user_role:  str     = data.body.model_dump(mode="json")["user"]["role"]
         # user_email: str     = data.body.model_dump(mode="json")["user"]["email"]
 
         chat_history_context: str = build_context_from_messages(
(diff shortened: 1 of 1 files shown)


backend/routers/cosmic.py
```diff
@@ -140,7 +140,7 @@ async def process_cosmic(
 ):
     try:
         user_id:    str     = data.body.model_dump(mode="json")["user"]["id"]
-        user_role:  str     = data.body.model_dump(mode="json")["user"]["role"]
+        # user_role:  str     = data.body.model_dump(mode="json")["user"]["role"]
```
carlosclaiton, line 143: my understanding is that here is the cosmic user, not the chat user and this would be required in case we want give any different capabilities for cosmic users, it is not the user in the chat conversation.

carlosclaiton (changes requested):
You are mixing thing, this will break as the cosmic require this user role:

Image

carlosclaiton (approved):
The problem I was facing during test seems not be related to this. 
You can merge this.

PR #163, opened 2026-07-18
Added separate fallback service. 

Adnan525:
Separated the fallback service and now it responds with what services are available.
image

### DIFF
--- a/api.py
+++ b/api.py
@@ -24,11 +24,11 @@ from .backend.routers import (
 # =====================Debugging========================
 # Uncomment the following lines to enable debugging:
 #
-# import debugpy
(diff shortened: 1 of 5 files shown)


carlosclaiton (commented):
Did you also check if the logic of the mapping function is working properly? When is this service activated?

Adnan525:
@carlosclaiton fixed copilot reviews and yes, mapping works well as it is

carlosclaiton (approved):
I have tested when 
- we have no enabled services, 
- with general questions only
- with academic governance and chess. 
- academic governance and general question

The behaviour is as expected.

PR #166, opened 2026-08-11
CoSMIC 331: Session clean

Adnan525:
The UI now sends a post request to CoSMIC with user ID and chat ID. We then use those to create the directory path where session files are usually located. subsequently, we clean up the document index JSON file and the vector database. Logs using print statements.


marslanshaukat:
Tested and its deleting vectors from VD and associated files from obj storage.

PR #170, opened 2026-09-03
RAG tiered retrieval [COSMIC-323]

marslanshaukat:
Implemented RAG tiered retrieval preventing context dilution across scopes and eliminating incorrect file attribution when answers originate outside attached files.

Also fixed Ollama response unpacking in qa.py, restoring QA execution and ensuring token usage is correctly tracked for main LLM calls.

### DIFF
--- a/src/services/qa.py
+++ b/src/services/qa.py
@@ -172,10 +172,18 @@ class QABase(ServiceBase):
             services=services # pyright: ignore[reportCallIssue]
         )
 
-        # The query analyser can route a question about a uploaded file into 0 or -1  
-        # RAG-eligible fallback branch so an attached file is considered
+        # The service the analyser actually chose (before any override below),
+        # for debugging.
+        analyser_service_option = service_option
+
+        # A file attached while the analyser was unsure (routed to system-info 0
+        # or fallback -1) still needs the retrieval branch so the file gets
+        # considered. Route it to the NEUTRAL general service (4) — not the
+        # academic-governance specialist (5) — so an unrelated file is not
+        # forced onto the governance knowledge base/persona. Fall back to 5 only
+        # if the general service is not registered.
         if has_files and service_option in ("0", "-1"):
-            service_option = "5"
+            service_option = "4" if "4" in services_name else "5"
 
         # Skip query as required or unknown service option.
         if query.find("skip") > -1:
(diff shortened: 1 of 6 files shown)


src/services/qa.py
```diff
@@ -277,7 +277,9 @@ def __call__(
             user_prompt = f"Select the best next move from {next_move} and explain why it is the best."
             (
```
FaceWithDark, line 348: I noticed about this bug while working on #161 as well. It supposed to be covered in that PR, hence what we just discussed in the meeting today.

We can safely ignore this as #167 is my attempt in fixing this.

src/services/rag.py
```diff
@@ -288,3 +291,65 @@ def _title(doc) -> str:
...
+        global_service_names: Optional[List[str]] = None,
+        document_ids: Optional[List[str]] = None,
+        include_user: bool = False,
+    ) -> tuple[str, list, str]:
+        """
+        Query memory tiers one at a time — session, then user (if enabled), then
+        global — stopping at the first tier whose result clears the relevance threshold.
+        """
+        # Tier 1: session/file scope.
+        context, scores = self(
```
carlosclaiton: it is not clear when this will be called?

Adnan525:
failed at the very first question? -
image
image

- please consider refactoring the __call__ in qa.py.
-  the comments are extremely confusing here: 
"        # A file attached while the analyser was unsure (routed to system-info 0
         # or fallback -1) still needs the retrieval branch so the file gets
         # considered. Route it to the NEUTRAL general service (4) — not the
         # academic-governance specialist (5) — so an unrelated file is not
         # forced onto the governance knowledge base/persona. Fall back to 5 only
         # if the general service is not registered.
         if has_files and service_option in ("0", "-1"):
             service_option = "4" if "4" in services_name else "5"",
since they all go to the same "else" block and the only difference that I could work out was "if answered_from_file and service_option in ("0", "-1", "4")"
- please don't import in the middle of the code:
"                from collections import Counter as _Counter
                print(
                    f"[qa] CONTEXT analyser_service={analyser_service_option!r}"
                    f"({services_name.get(analyser_service_option, '')}) "
                    f"routed_service={service_option!r}({services_name.get(service_option, '')}) "
                    f"persona={answering_service!r} "
                    f"routed_to_kb={routed_to_kb} "
                    f"file_directed={file_directed} docs={document_ids or None} "
                    f"tiers={dict(_Counter(s['tier'] for s in retrieved_sources))} "
                    f"scores={[round(s['score'], 4) for s in retrieved_sources]} "
                    f"ctx_chars={len(context_retrieved)} q={query[:60]!r}"
                )"
- "        document_ids: List[str] = []
        attached_file_names: List[str] = []" -- Using "List" without importing will trigger an error
- "        # Which knowledge base (if any) does this query want? Use the query
        # analyser's own routing decision, not keyword matching: the analyser
        # is an intent classifier that already picked the service this query is
        # about, and each service carries a `memory_capability` flag (surfaced
        # here as membership of `global_service_names`, the active
        # memory-capable services)." - i am a bit confused with those comments, just a few lines before, we have a list called "FILE_INTENT_MARKERS" that does key word matching.

Adnan525 (changes requested):
please see my comment above

marslanshaukat:
> please see my comment above

image

Thank you for your feedback.

I have retested the setup, and it is now working correctly. I followed the steps below:

1. Cleaned the environment using docker system prune and removed the Docker volumes.
2. Deleted the CoSMIC repository and cloned it again from scratch.

Regarding your comment about the imports appearing in the middle of the code, this is intentional. These imports are related to logging, and I plan to use them for tracing and observability purposes in the upcoming PR. If you review the CoSMIC logs and execute a query from the frontend, you will be able to see the complete execution path, including:

Which service was triggered by the query analyser
Which chunks were retrieved through dense retrieval
Which layer the retrieved chunks originated from

I would also recommend reviewing [COSMIC-323.pdf](https://github.com/user-attachments/files/32060096/COSMIC-323.pdf) that provides a detailed explanation of things.

Please let me know if you encounter any issues while reproducing the results. Thanks.

Adnan525:
Hi Arslan. The comments still have to be fixed, including a couple of points that I raised earlier. Please see below:
- I have gone through the pdf several times and understood the architecture we are aiming for including the complications it brings in. Nevertheless, the implementation does no justice to this architecture. It's riddled with confusing comments, long and inconsistent codebase.
- Regardless of intent of use, the import statements should go to the top. Doesn't help that the rag.py file have several of these imports in the middle of the codebase even though they are not used for logging.
image

- The doc string formatting is not consistent. Type annotation is also not consistent e.g. some of the functions have a return type, some don't, some have no type annotation. I can see that a few of them are not from this PR but they are all from your PRs that were previously merged.
image
image
image
image

- The qa.py has to be refactored. Please remove all unnecessary comments, refactor this code so we use single-intent functions. Create a new helper class/service if required
- Double check this - " document_ids: List[str] = []
attached_file_names: List[str] = []" -- Using "List" without importing will trigger an error
- Also please break down this ticket into sub tasks. since this is doing a lot of things - e.g. reranking, intent detection, detect attached file etc; it's extremely hard to keep track of all changes, especially when code quality poor. **I will be closing this PR, please first clean up the code base and then create sub tickets for this ticket.** Thanks.

PR #171, opened 2026-09-16
Replaced hard-coded RAG eligibility [COSMIC-329]

martin213214:
replaced the hard coded RAG eligibility


src/services/qa.py
```diff
@@ -175,7 +175,12 @@ def __call__(
         # The query analyser can route a question about a uploaded file into 0 or -1  
         # RAG-eligible fallback branch so an attached file is considered
         if has_files and service_option in ("0", "-1"):
-            service_option = "5"
```
marslanshaukat, line 178: This is meant to be like this as the logic behind is to fall back if query analyser suggest a service 0 and 1 but there was file attached to it. We do not want to change this functionality yet as it will be resolved one query analyser is fixed.

docker-compose.yaml
marslanshaukat, line 1: I understand changes have been made to run on your setup but you don't need to add and commit those changes.

compose.yaml
marslanshaukat, line 1: I understand changes have been made to run on your setup but you don't need to add and commit those changes.

martin213214:
i made the changes as requested

marslanshaukat:
Tested and verified successfully across all repo's (CoSMIC, CoSMIC_DB, CoSMIC_UI), no issue in their working.

